PR #9, opened 2015-09-18
Fix compilation with BOOST_NO_ANSI_APIS

MarcelRaad:
With BOOST_NO_ANSI_APIS, there is no CryptAcquireContextA.

### DIFF
--- a/include/boost/uuid/seed_rng.hpp
+++ b/include/boost/uuid/seed_rng.hpp
@@ -92,7 +92,7 @@ public:
         , random_(NULL)
     {
 #if defined(BOOST_WINDOWS)
-        if (!boost::detail::winapi::CryptAcquireContextA(
+        if (!boost::detail::winapi::CryptAcquireContextW(
                     &random_,
                     NULL,
                     NULL,


mclow:
I have no opinion on this change. It looks simple and straightforward, but I can't say about correctness (not knowing much about windows programming)

Lastique:
It does look incorrect because the `crypt_acquire_context` call is ambiguous when `BOOST_NO_ANSI_APIS` is not defined.

MarcelRaad:
Thanks @Lastique, fixed now. This is the version I had originally against Boost 1.59, but I thought I could avoid the ifdef with the recent winapi develop changes and forgot to re-test without  BOOST_NO_ANSI_APIS :-(

Lastique:
You can avoid the preprocessor check if you just always call `CryptAcquireContextW`.

MarcelRaad:
Ah, of course, thanks!

MarcelRaad:
Ping?

MarcelRaad:
Thanks!

PR #10, opened 2015-09-30
Make seed_rng compatible with MinGW

pkerling:
See https://svn.boost.org/trac/boost/ticket/11693


eldiener:
This needs to be merged into 'develop', not 'master'.

pkerling:
I see that this is already fixed in develop.

PR #12, opened 2015-12-29
Fixed padded output of UUID when std::left is set

erichkeane:
The defect https://svn.boost.org/trac/boost/ticket/11385 reports a
situation where a uuid's single-hex-digit values are improperly padded
when std::left is set on the stream.

It is fixed by ensuring that the ostream overload always uses std::right
to be consistent with the default behavior.

### DIFF
--- a/include/boost/uuid/uuid_io.hpp
+++ b/include/boost/uuid/uuid_io.hpp
@@ -46,7 +46,7 @@ template <typename ch, typename char_traits>
             }
         }
 
-        os << std::hex;
+        os << std::hex << std::right;
         os.fill(os.widen('0'));
 
         std::size_t i=0;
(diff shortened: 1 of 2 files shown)


mclow:
Needs a test.

erichkeane:
Thanks for the feedback @mclow.  I believe I added a satisfactory test to the test_io in commit 2449c17.

mclow:
Sorry; somehow I missed that.  I am concerned that you're changing the alignment of the stream, which may lead to confused users. Of course, the code already does that for `std::hex`.

Maybe this should use http://www.boost.org/doc/libs/1_60_0/libs/io/doc/ios_state.html to save and restore the state of the stream. What do you think?

erichkeane:
@mclow It seems that lines 34 and 35 of uuid_io.hpp already use the flags/fill saver to do that.  I tested in a sample app (though perhaps I should add a unit test) to ensure that the stream-alignment is restored.

Or does the state_saver do something different from the flags_saver that I'm missing?

mclow:
That will teach me to look at the whole file, rather than the just the diff. Thanks for your patience.

erichkeane:
Great, thanks!  I appreciate the comments, it definitely got me double checking everything.

PR #15, opened 2016-04-21
Fixed missing boost:: namespace

aldonin:
Fixed missing  ```boost::``` namespace

### DIFF
--- a/test/test_wserialization.cpp
+++ b/test/test_wserialization.cpp
@@ -67,5 +67,5 @@ int test_main( int /* argc */, char* /* argv */[] )
     test_archive<xml_woarchive, xml_wiarchive, wostringstream, wistringstream>();
     test_archive<binary_woarchive, binary_wiarchive, wostringstream, wistringstream>();
 
-    return report_errors();
+    return boost::report_errors();
 }


mclow:
This looks fine to me, but I wonder how this ever worked.  Can you point to a test case that fails before/succeeds after?

aldonin:
Hello @mclow, 
I don't actually understand what do you want to see? 
Seems like it began after [this](https://github.com/boostorg/uuid/commit/3ca9d3fae00805833d42210e51f24966939fbc05) commit. Before it, the previos version of [test_wserialization.cpp](https://github.com/boostorg/uuid/blob/3ca9d3fae00805833d42210e51f24966939fbc05/test/test_wserialization.cpp) ended with:
```return 0;```
But commit changed this to:
```return report_errors();```
So as you can see the most part of another tests ended with ``` return boost::report_errors();```
But in our case this `boost::` namespace is missing. So in case of compiling I get error message 

> use of undeclared identifier 'report_errors'

And adding ```boost::``` namespace fix this problem. 

I hope I clarify this moment for you.
Thanks in advance!

PR #16, opened 2017-01-11
Use null_deleter from Boost.Core instead of own definition

tklauser:


mclow:
I think that this is a tossup.
On one hand, it removes a (small) class from Boost.UUID.
On the other hand, it adds a dependency on Boost.Core.
On the gripping hand, most everything in boost already depends on Boost.Core.

tklauser:
boost/uuid/seed_rng.hpp already has a dependency on a header from Boost.Core (boost/core/noncopyable.hpp), so I thought it should be fine add another dependency. I probably should have mentioned this in the PR, sorry.

mclow:
Please keep an eye on the test bots, and ping me in a week or so to merge to master.

tklauser:
Will do. Thanks!

tklauser:
@mclow pinging for merge to master. The test bots didn't show any failure which I could trace to this change.

tklauser:
@mclow pinging again for merge to master

mclow:
This has been done for the 1.64.0 release.

PR #17, opened 2017-08-03
Fix boost trac 8630, 9343, 11482, 11483, 12253 improving invalid string generator handling in uuid

jeking3:
For example, "12b672ea-388d-47da-9c26-bec5bc8a06cb-foo" was not considered invalid before, but it is now.

### DIFF
--- a/include/boost/uuid/string_generator.hpp
+++ b/include/boost/uuid/string_generator.hpp
@@ -81,15 +81,21 @@ struct string_generator {
                 }
             }
             
-            if (has_dashes) {
-                if (i == 6 || i == 8 || i == 10) {
+            // if there are dashes, they must be in every slot
+            else if (i == 6 || i == 8 || i == 10) {
+                if (has_dashes == true) {
                     if (is_dash(c)) {
                         c = get_next_char(begin, end);
                     } else {
                         throw_invalid();
                     }
+                } else {
+                    if (is_dash(c)) {
+                        throw_invalid();
+                    }
                 }
             }
+                
 
             *it_byte = get_value(c);
 
@@ -102,6 +108,11 @@ struct string_generator {
         if (has_open_brace) {
             c = get_next_char(begin, end);
             check_close_brace(c, open_brace_char);
+        }
+
+        // check end of string - any additional data is an invalid uuid
+        if (begin != end) {
+            throw_invalid();
         }
         
         return u;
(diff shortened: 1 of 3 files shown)


eldiener:
I would like to see tests for all 4 forms when generating from a string and there is unexpected extra data at the end of the string. Your test is good for 1 of the 4 different forms.

jeking3:
@eldiener I updated the pull request with the additional tests.  If I can be of any assistance in grooming the boost trac backlog, let me know.

include/boost/uuid/string_generator.hpp
```diff
@@ -81,15 +81,21 @@ struct string_generator {
                 }
             }
             
-            if (has_dashes) {
-                if (i == 6 || i == 8 || i == 10) {
+            // if there are dashes, they must be in every slot
+            if (i == 6 || i == 8 || i == 10) {
```
eldiener: Can this not be:
`else if (i == 6 || i == 8 || i == 10) {`

include/boost/uuid/string_generator.hpp
```diff
@@ -122,23 +133,27 @@ struct string_generator {
         static char const*const digits_end = digits_begin + 22;
```
eldiener: This isn't your code but can't be take the sizeof(digits_begin) and use that instead of hardcoding 22 and further on down hardcoding '> 21' ? Hardcoding numbers is generally poor c++ if there is a better way to get the hardcoded value from the compiler.

eldiener:
The code works. Please see my comments for some possible individual changes. 

I think the docs should be updated to both explain all the valid string formats uuid accepts and specifies that uuid throws a particular exception if the string is not a valid format or contains invalid hex digits.

jeking3:
I wonder if a boost::regex would be sufficient not only to describe the valid format but also enforce and to parse it, in the future.  That would force a dependency that folks may not want however, and it may not be documentation-friendly.

include/boost/uuid/string_generator.hpp
jeking3: sizeof(digits_begin) if changed to ``static wchar_t const digits_begin[] = ...`` is 46, confirmed by a BOOST_STATIC_ASSERT, so in this case I think the hardcoded values is appropriate.  Considering the enhanced testing, any changes here will be caught.

include/boost/uuid/string_generator.hpp
jeking3: It can, I will push a change for it.

jeking3:
I pushed an update that addresses all concerns thus far.

include/boost/uuid/string_generator.hpp
eldiener: In the get_value(char c) case it should be:
`sizeof(digits_begin)/sizeof(char)`
and in the get_value(wchar_t c) case it should be:
`sizeof(digits_begin)/sizeof(wchar_t)`

so that hardcoded numbers are not used.

jeking3: Okay, we'll have to subtract one, as the digits_end needs to point at the trailing NULL, but I will make this change.

PR #19, opened 2017-08-04
9343: fix incorrect uuid format in string generator comment

jeking3:


eldiener:
Please add this change to the PR you already have open and close this PR. I am looking at your other PR and would rather not have two separate PRs encompassing the same issue(s).

jeking3:
@eldiener that's fine, sometimes folks want each fix to be distinct and separate, and this could be separated from the others.  I'll roll it into the changes on the other.

jeking3:
This never got rolled into the other documentation pull request so I am re-opening it so it can be merged.

PR #20, opened 2017-08-10
12280: fix documentation file locations to conform with boost library requirements

jeking3:
From 12280:

I would recommend that we follow the pattern used in the boost/thread module, where boost/uuid/index.html would become a redirect to doc/index.html, and both index.html and uuid.html would move into boost/uuid/doc. Would this be sufficient to bring it in line with the style of the other modules?

This is what I did.

@grafikrobot would you be able to review this change so we know it meets the requirements you put into Trac 12280?  Thanks.


jeking3:
@grafikrobot is the same change needed in boost::format?  If so I can add an issue to it.  Let me know.

PR #21, opened 2017-08-10
12258: re-enable unit test for random_device generator

jeking3:
Investigating Trac 12258 I found the random_device unit test was not enabled.  This establishes a formal dependency during the build of uuid on random, but that seems quite appropriate.  Tested on Windows (MSVC2015) and Linux (gcc-4.8, trusty).

### DIFF
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -34,7 +34,7 @@ test-suite uuid :
     [ run test_nil_generator.cpp ]
     [ run test_name_generator.cpp ]
     [ run test_string_generator.cpp ]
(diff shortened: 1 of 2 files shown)


eldiener:
Did you check the history for test_random_generator.cpp to see why it was disabled ? I agree with you that it should not have been disabled simply because the test establishes a dependency on another Boost library.

jeking3:
A fine idea, I will do that!

jeking3:
It was always disabled, as far back as 8 years ago.  No history on why it
was commented out.

Issue #24, opened 2017-09-04
seed_rng.hpp does not compile for Windows UWP apps


strentler:
These Win32 functions are not allowed to be used with UWP apps:

CryptAcquireContext
CryptReleaseContext
CryptGenRandom

see here
https://msdn.microsoft.com/en-us/library/windows/desktop/aa379886(v=vs.85).aspx

Some replacement functions should be used instead.

I have tried to compile using this Windows SDK/ Target Platform Version: 10.0.15063.0

jeking3:
This sounds similar to the following:

https://svn.boost.org/trac10/ticket/11841

Could you try the workaround provided there and see if it works for you?
I'm considering removing all the built-in random in boost UUID, and requiring the use of boost::random to initialize random behavior.

eldiener:
Removing built-in random, if you do, would be a breaking change, so please be aware of the usual cycle of deprecating at least one release first before eventually changing.

jeking3:
Sounds good, I'll have to look at what it is doing and make sure it is okay to deprecate it.  In the meantime, hopefully the workaround provided helps @strentler.

jeking3:
@eldiener perhaps the default constructor could internally switch to using something from boost::random which would allow the internal random/crypt code to be removed without breaking the external interface.  I'll take a look...

jeking3:
@eldiener the public documentation for basic_random_generator makes no guarantees or mention of using the built-in seed mechanism.  Given mt19937 seeds itself properly (if it didn't there would be lots of other problems in lots of other code), it may be simple enough to change the default constructor not to call detail::seed and remove seed_rng.cpp since it really isn't needed.  The question is whether this becomes a breaking change.  Given the constructor is not specifying a seed, and given any UniformRandomNumberGenerator must properly seed itself to be useful, do you see anything that needs to be deprecated here, or can we just remove the unnecessary seeding implementation here and just rely on boost::random to do the right thing?

swatanabe:
AMDG

  A default constructed mt19937 will always produce the
same sequence, which isn't desirable for uuid.

In Christ,
Steven Watanabe

jeking3:
Thanks, I submitted a pull request today that should address this by
leveraging boost::random::random_device to provide the entropy in a more
platform-friendly way.

jeking3:
Anyone know how to "connect" this issue to the PR #25?

Lastique:
[...]

PR #25, opened 2017-09-04
#24: improve platform compatibility by changing default random seed

jeking3:
from uuid internal code which was not friendly to all platforms, to boost::random::random_device which is
I'm submitting this PR for comments.

@eldiener this requires a new dependency for some use cases; folks may require linking against boost::random if they are using boost::uuid and relying on the default random generator constructor whereas before the code was platform specific and not always compatible.

Summary:

1. Moved seed_rng into detail/
2. Removed all platform-specific logic in seed_rng and changed it to leverage boost::random::random_device.
3. Updated documentation and comments appropriate to change.

Should consider moving seed_rng into boost::random at some point to make it easier to use seed_seq, or fastly simplify boost::random seeding of PRNGs by allowing each PRNG to take a boost::random::random_device as an argument to seed().


eldiener:
Why should just changing the order of the header files in random_generator.hpp ever matter ?

jeking3:
While header order should not (and in this case does not) matter, I prefer
to have headers ordered alphabetically with the exception of the header for
the module it is implementing which belongs first.

include/boost/uuid/detail/seed_rng.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace boost {
+namespace uuids {
+namespace detail {
+
+// seed() seeds a random number generator with good seed values
+template <typename UniformRandomNumberGenerator>
+inline void seed(UniformRandomNumberGenerator& rng)
+{
+    BOOST_CONSTEXPR_OR_CONST size_t entropy_len = 16; // 512 bytes of entropy
```
jeking3: Should boost/config.hpp be added as an include to satisfy BOOST_CONSTEXPR_OR_CONST usage here?

strentler:
@jeking3 : i have applied your patch. It does the job for UWP now.  Have tested this with SDK version 10.0.15063.0. I have also checked if it goes through the "Windows App Cert Kit".  It does.
Thank you!

jeking3:
@eldiener or anyone else on the CTP team able to do a code review?

include/boost/uuid/detail/seed_rng.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace uuids {
+namespace detail {
+
+// seed() seeds a random number generator with good seed values
+template <typename UniformRandomNumberGenerator>
+inline void seed(UniformRandomNumberGenerator& rng)
+{
+    BOOST_STATIC_CONSTEXPR size_t entropy_len = 32; // 256 bits of entropy
+    boost::random::random_device rd;
+    std::vector<unsigned int> entropy(entropy_len);
```
Lastique: This should be an array if the size is known in compile time (it is in your case).

include/boost/uuid/detail/seed_rng.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace boost {
+namespace uuids {
+namespace detail {
+
+// seed() seeds a random number generator with good seed values
+template <typename UniformRandomNumberGenerator>
+inline void seed(UniformRandomNumberGenerator& rng)
+{
+    BOOST_STATIC_CONSTEXPR size_t entropy_len = 32; // 256 bits of entropy
```
Lastique: `std::size_t`. Also, `#include <cstddef>` is missing.

include/boost/uuid/detail/seed_rng.hpp
```diff
@@ -0,0 +1,47 @@
...
+    boost::random::random_device rd;
+    std::vector<unsigned int> entropy(entropy_len);
+    std::generate(entropy.begin(), entropy.end(), std::ref(rd));
+    boost::random::seed_seq seq(entropy.begin(), entropy.end());
+    rng.seed(seq);
+}
+
+// random_device itself does not / can not be seeded
+template <>
+inline void seed<boost::random::random_device>(boost::random::random_device&) {}
```
Lastique: This should probably be an overload instead of specialization.

Lastique:
I would rather keep `seed_rng` so that it is possible to generate UUIDs without linking with Boost.Random. `seed_rng` uses system sources for getting random data that are ultimately used by Boost.Random, so there's little reason to bring in that dependency. I would rather have Boost.Random-based version separate, if possible, and without breaking backward compatibility.

jeking3:
@Lastique my rationale for eliminating the platform-specific code in seed_rng was to avoid situations such as that described in #24.  Given there is already a documented, supported, and highly portable seed generator in boost::random::random_device, it makes less sense to me to have a similar method in uuid with platform specific code that needs to be maintained.  Could you explain further how you believe the change I made is not backwards compatible?  seed_rng is in the detail namespace and it is my understanding that these do not need to maintain backwards compatibility.  

As with all projects, I am looking to minimize the maintenance cost of the library by leveraging that which already exists and likely does a far better job than the code that I removed.  It should be intuitive from a user perspective that if they are using random_generator to make randomized UUIDs, that they would need to link to boost::random.

Thank you for your other code review comments, I will make changes to address them.

Lastique:
On 09/07/17 20:00, James E. King, III wrote:
> my rationale for eliminating the 
> platform-specific code in seed_rng was to avoid situations such as that 
> described in #24 <https://github.com/boostorg/uuid/issues/24>. Given 
> there is already a documented, supported, and highly portable seed 
> generator in boost::random::random_device, it makes less sense to me to 
> have a similar method in uuid with platform specific code that needs to 
> be maintained.

I don't think the cost of maintenance here is significant enough when 
compared to having Boost.UUID a header-only library. In some projects, 
header-only property is a deal breaker.

The #24 needs to be investigated. I'm not familiar with the UWP 
specifics, but I'm sure there must be an API for accessing system RNG.

> Could you explain further how you believe the change I 
> made is not backwards compatible? seed_rng is in the detail namespace 
> and it is my understanding that these do not need to maintain backwards 
> compatibility.

Having to link with Boost.Random is a breaking change.

> It should be intuitive from 
> a user perspective that if they are using random_generator to make 
> randomized UUIDs, that they would need to link to boost::random.

No, I'm sorry, that doesn't follow. Boost.Random is a separate library 
and Boost.UUID is not obliged to use it if there is enough reason not 
to. IMHO, header-only property is important enough to avoid the linking 
dependency on Boost.Random, especially since the seeding code is already 
written and works on most platforms (except UWP, that is). Adding 
support for another platform to the existing implementation seems more 
preferable to me.

jeking3:
@Lastique your argument for header-only is convincing enough for me to reconsider.  I do wonder though if seed_rng belongs in boost::random more than in uuid, since there will be subject-matter experts there who will be able to deal with the issue of platform compatibility should it pop up again in the future.  For the time being I will close out this change and focus on fixing the specific bug reported.

Lastique:
> I do wonder though if seed_rng belongs in boost::random more than in uuid

This is probably true. `seed_rng` wouldn't be needed if `boost::random_device` was header-only.

PR #33, opened 2017-09-10
Promote recent changes to master

jeking3:
CI builds of this merge: https://github.com/jeking3/uuid/pull/2


mclow:
Usually what we do is to request someone to merge a previous PR (or set of commits) (or all the changes) from `develop` to `master`.  We don't do pull requests for `master`.

This *appears* to be all the changes between `develop` and `master` - is that correct?

jeking3:
Correct, however I like to use a pull request to validate the merge through CI which is what the PR in my own account (in the description) is and it built clean.  In the future I can do this on my own account, wait for the results, and promote it without a PR.  I'll close this and do that, does that sound ok?

jeking3:
Closing - this was done only to get a CI build set for verification before promoting to master.

mclow:
What I usually do is watch the testbots results; that way you get to see a bunch of compilers/OSes, not just one.

jeking3:
@mclow I'll leverage that in the future.  There seem to be a few errors in other modules showing up here however, and I'm curious why one would build with clang++-4.0 using libstdc++-4.8 (because the latter is quite old) - for example in this result:

mclow:
> I'm curious why one would build with clang++-4.0 using libstdc++-4.8
I agree, it's weird - but we've got lots of users, some with odd set-ups. Some of them have good reasons for them. (I don't know if this is one of the ones with a good reason)

PR #36, opened 2017-09-25
refactor name_generator to have a configurable hash algorithm

### LINKED ISSUE #26: name_generator uses a fixed, insecure hash algorithm (SHA1)
Earlier this year Google announced that they had produced the first SHA1 collision:

https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html

The boost::uuid library maintains a SHA1 implementation today for purposes of the name_generator, which takes any arbitrary sized set of data and generates a UUID from it.  The library is hardcoded to use the SHA1 hash algorithm today.

The proposed improvement here is to make the hash algorithm configurable.  A new HashingProvider concept would be introduced (or at least discussed) and/or a pure virtual base specifying the required methods an algorithm would need to implement (process_byte, process_bytes, get_digest); modify name_generator to be a template class where the hash algorithm is supplied as a template parameter, and checked to be a base of this virtual class at compile time such that a usable error message would occur.

To maintain backwards compatibility a default template parameter leveraging the existing SHA1 hash algorithm would be introduced, however it would be deprecated, and eventually the default would be removed thus forcing all use of name_generator to specify the hash algorithm as a mechanism to gently move them away from the less secure SHA1 to something else.

It may also be useful to move hashing to its own library where the HashingProvider concept could live, similar to how the UniformRandomNumberGenerator concept lives in boost::random.

jeking3:
and provide backwards compatibility for sha1, and also added
md5 to complete the RFC 4122 spec implementation

as such, boost::uuids::name_generator is deprecated, however
still defined to use the same implementation in previous boost
releases, and name_generator_sha1 as well as name_generator_md5
now exist, in preparation for whatever will replace sha1.

to properly test the new header structure, I took the bjam
rule from winapi to isolate each header as an include and make sure it
has no dependencies on other headers to be included first,
and was able to remove a few test files that became unnecessary.

Appveyor: https://ci.appveyor.com/project/jeking3/uuid/build/1.0.59-develop
Travis: https://travis-ci.org/jeking3/uuid/builds/280110745

This fixes #26


include/boost/uuid/name_generator.hpp
```diff
@@ -1,21 +1,27 @@
...
 
 #ifndef BOOST_UUID_NAME_GENERATOR_HPP
 #define BOOST_UUID_NAME_GENERATOR_HPP
 
-#include <boost/uuid/uuid.hpp>
-#include <boost/uuid/sha1.hpp>
+#include <boost/concept_check.hpp>
+#include <boost/cstdint.hpp>
 #include <boost/static_assert.hpp>
+#include <boost/uuid/detail/md5.hpp>
```
Lastique: Please, avoid including unnecessary code. I would suggest moving `basic_name_generator` to a separate (new) header and leaving only `name_generator` in name_generator.hpp. The other typedefs with their respective includes can go in their new headers.

Another way to do this is create a name_generator directory with the headers, each header defining one component, and making name_generator.hpp include all headers from that directory. The headers in the name_generator directory can be named after the components they define.

The point is to allow users to include only the bits they need and avoid the rest. No one will use the MD5-based generator, so this is dead code 99% of the time and only slows down compilation and adds potential problems with compilation.

include/boost/uuid/sha1.hpp
```diff
@@ -1,231 +1,16 @@
...
-#include <boost/throw_exception.hpp>
-#include <cstddef>
-#include <string>
-
-#ifdef BOOST_NO_STDC_NAMESPACE
-namespace std {
-    using ::size_t;
-} // namespace std
+#if defined(__GNUC__) || defined(_MSC_VER)
+#pragma message("This header is deprecated, use boost/uuid/detail/sha1.hpp instead.")
```
Lastique: I think, the message should say that the header is implementation detail.

doc/uuid.html
```diff
@@ -672,6 +700,44 @@
...
+<tr>
+	<td><tt>void G::process_byte(unsigned char)</tt></td>
+	<td></td>
+</tr>
+<tr>
+	<td><tt>void G::process_bytes(void const*, std::size_t)</tt></td>
+	<td></td>
+</tr>
+<tr>
+	<td><tt>void G::get_digest(typename G::digest_type&)</tt></td>
```
Lastique: Replace `&` with `&amp;`.

Lastique:
On 09/26/17 04:22, James E. King, III wrote:
> @Lastique I'd appreciate if you could look 
> over my use of Boost.Concept here I'm not sure it's really necessary. I 
> added the definition into the documentation as well and I updated

I didn't use Boost.Concept, so I'm not sure I can comment on the use of it. I do think though that it is better to avoid the overhead from using it, if possible. I wouldn't mind if the concept was defined in a separate header that is not included and not used by default but can be enabled upon user's request. I'm not sure if Boost.Concept offers any tools to that effect.

jeking3:
@Lastique my guess is I'm not using concept well here, so I will probably remove it since it is documented.  I found all it does is double the compile failures when an implementation is missing something, and the only value-add I could see is forcing const on get_version which is not terribly important.

include/boost/uuid/sha1.hpp
jeking3: Given sha1 and seed_rng are already in the detail namespace, am I required to provide this mechanism?  I believe the answer is no, but I want to make the transition smooth because I suspect there are many projects that rely on uuid sha1 just for the sha1 implementation.

Lastique: I don't think relying on Boost.UUID just for the SHA1 implementation is reasonable. Especially given that it is an implementation detail. That said, the header is in the public directory, so I still would give it a deprecation period. My comment was just about the message, which currently suggests to continue using the header at the new location while I think it should make clear that people shouldn't be using it in the first place.

jeking3: I found that,

Boost.Compute is including uuid/sha1: ``compute/include/boost/compute/detail/sha1.hpp``
Boost.Beast has another sha1 implementation: ``beast/include/boost/beast/core/detail/sha1.hpp``

Usually when there is more than one available, it's time to start looking into consolidating.  This is why I asked the question about something like Boost.Hash or Boost.Crypto to hold commonly used things like this.  The pushback was that area usually has a lot of specialized functions and it would be better to leave it out.  So for now I guess we have two SHA1 implementations.  Boost.Compute technically needs one of its own so it does not rely on uuid/detail, or we make a more common library for it.

jeking3:
I believe I addressed all your concerns from the previous review.

include/boost/uuid/detail/md5.hpp
```diff
@@ -0,0 +1,340 @@
...
+    static MD5_u32plus F(MD5_u32plus x, MD5_u32plus y, MD5_u32plus z) { return ((z) ^ ((x) & ((y) ^ (z)))); }
+    static MD5_u32plus G(MD5_u32plus x, MD5_u32plus y, MD5_u32plus z) { return ((y) ^ ((z) & ((x) ^ (y)))); }
+    static MD5_u32plus H(MD5_u32plus x, MD5_u32plus y, MD5_u32plus z) { return (((x) ^ (y)) ^ (z)); }
+    static MD5_u32plus H2(MD5_u32plus x, MD5_u32plus y, MD5_u32plus z) { return ((x) ^ ((y) ^ (z))); }
+    static MD5_u32plus I(MD5_u32plus x, MD5_u32plus y, MD5_u32plus z) { return ((y) ^ ((x) | ~(z))); }
+
+    /*
+     * The MD5 transformation for all four rounds.
+     */
+    #define __MD5_STEP(f, a, b, c, d, x, t, s) \
```
Lastique: Please, rename the macro to follow Boost conventions. No leading underscores, `BOOST_UUID_DETAIL_` prefix in the name to not clash with anything else. Ditto for other macros below.

jeking3: Sure thing.

jeking3: I'll take care of the static private methods in there too.

jeking3:
Cleaned that up.  I tried turning on ``-Wextra -Werror -pedantic`` but it uncovered a few more things than I wanted to include here, so I'll defer that to another issue.

jeking3:
Note I still have to use my github account to run the CI builds because they haven't been turned on here yet.

jeking3:
This one is all set for another round of review.

include/boost/uuid/detail/md5.hpp
```diff
@@ -0,0 +1,341 @@
...
+    } MD5_CTX;
+
+    /*
+     * The basic MD5 functions.
+     *
+     * F and G are optimized compared to their RFC 1321 definitions for
+     * architectures that lack an AND-NOT instruction, just like in Colin Plumb's
+     * implementation.
+     */
+    BOOST_FORCEINLINE MD5_u32plus F(MD5_u32plus x, MD5_u32plus y, MD5_u32plus z) { return ((z) ^ ((x) & ((y) ^ (z)))); }
```
Lastique: I thought you wanted to rename one or two caps letter long functions like this one to avoid clashes with macros, did you not?

jeking3: Sure I can do that too.

jeking3:
All set for another round of review.  Those weren't macros (F, G, H, H2, I) as I had already converted them to inline functions.  As such, being in the same compilation unit they would take precedence, but I renamed them to use the longer form anyway.

jeking3:
@Lastique given you previously marked this as "request changes", Id' appreciate it if you could give it one final lookover before I merge it.  Thanks.

PR #45, opened 2017-10-14
resolve some build warnings identified in the test matrix

jeking3:


include/boost/uuid/uuid_io.hpp
```diff
@@ -20,11 +20,6 @@
 #include <locale>
 #include <algorithm>
 
-#if defined(_MSC_VER)
-#pragma warning(push) // Save warning settings.
-#pragma warning(disable : 4996) // Disable deprecated std::ctype<char>::widen, std::copy
```
Lastique: By removing this code you enable warnings in the users' code.

jeking3: This was done intentionally.  I receive zero warnings on msvc-10.0 through 14.1 in the unit tests for uuid (I built with warnings-as-errors to be sure).  Perhaps there is a missing unit test.

I see in other libraries such as random, it is up to the build command (see random/Jamfile.v2) to disable these warnings by defining `_SCL_SECURE_NO_WARNINGS` and not up to the code.

If the regression matrix for uuid shows no warnings, do we still need to disable 4996?

Lastique: Disabling warnings in the Jamfile is only good when building the library (i.e. it is beneficial only when the warning is triggered by the source code that is never included by users). In headers the warnings should be suppressed in code.

I cannot test MSVC now, so I don't know if the warning is triggered by tests on some other compiler version. But it seems not impossible that the tests don't trigger the warning. I can see both `widen` and `std::copy` are still used in the code.

jeking3: My point being anybody compiling with Boost.Random will need to do the same thing to disable the warnings as well, so I was trying to be uniform.  I see no pragmas to disable 4996 in random code anywhere.  I suspect this is intentional and it is up to the consuming build to deal with it?

I can put this back though.  It doesn't hurt anything to leave it there.  It just didn't seem like it was necessary any more based on the (likely incomplete) tests.

PR #53, opened 2017-11-08
Refactoring of random provider

### LINKED ISSUE #24: seed_rng.hpp does not compile for Windows UWP apps

jeking3:
Changed the default random_generator implementation to use
operating-system provided entropy as it is more secure and
faster for the typical use case of generating one uuid at
a time.

This is a breaking change for anyone passing a mt19937
into one of the explicit constructors of random_generator,
which would be quite rare.

Changed the default random provider on Windows to use BCrypt
where available, falling back to Wincrypt when necessary or
when explicitly requested through a macro.

Provide a new random_generator_mt19937 type definition for
use cases where a large number of uuids need to be created
with high performance.  This is equivalent to the previous
definition of random_generator.

Provide a random generation benchmark test showing the
cutoff where the mt19937-based generator will outperform the
standard generator based on wall time.

Removed template specialization for boost::random::random_device
so that any UniformRandomNumberGenerator can be used properly
with random_generator.

Replaced the seed_rng detail implementation (which had a number
of flaws) with a replacement header-only random_provider
implementation.

Note: entropy generation errors will cause an entropy_error
to be thrown from random_generator.  The previous implementation
ignored errors and silently failed.

Added internal support for entropy generation on cloudabi
platform leveraging the new random_device implementation.

Added internal support for Universal Windows Platform (UWP)
development leveraging the new random_device implementation.

Added internal support for getentropy() on Linux and OpenBSD
if certain requirements are met.

This is an evolution of #52
This fixes #24


include/boost/uuid/detail/random_provider.hpp
```diff
@@ -0,0 +1,63 @@
...
+      public noncopyable
+{
+public:
+    //! Leverage the provider as a SeedSeq for
+    //! PseudoRandomNumberGeneration seeing.
+    //! \note: See Boost.Random documentation for more details
+    template<class Iter>
+    void generate(Iter first, Iter last)
+    {
+        BOOST_STATIC_ASSERT(sizeof(Iter) >= sizeof(unsigned int));
```
pdimov: This checks the iterator size, which is irrelevant. It should check

- whether the `value_type` of the iterator is integral;
- whether the `value_type` of the iterator is unsigned;
- whether `numeric_limits<value_type>::max()` is >= 2^32-1, but this can only be tested at compile time on C++11, so
- whether `sizeof(value_type) * CHAR_BIT >= 32`.

include/boost/uuid/detail/random_provider.hpp
```diff
@@ -0,0 +1,63 @@
...
+        for (; first != last; ++first)
+        {
+            get_random_bytes(&*first, sizeof(*first));
+            *first &= (std::numeric_limits<unsigned int>::max)();
```
pdimov: Not correct. `generate` is required to produce numbers in [0..2^32-1], regardless of the size of `unsigned int`.

include/boost/uuid/detail/random_provider.hpp
```diff
@@ -0,0 +1,63 @@
...
+#define BOOST_UUID_DETAIL_RANDOM_PROVIDER_HPP
+
+#include <boost/core/noncopyable.hpp>
+#include <boost/cstdint.hpp>
+#include <boost/limits.hpp>
+#include <boost/static_assert.hpp>
+#include <boost/uuid/entropy_error.hpp>
+
+#include <boost/uuid/detail/random_provider_detect_platform.hpp>
+#include BOOST_UUID_RANDOM_PROVIDER_BASE_INCLUDE()
```
pdimov: It's better to just include the appropriate platform header from `random_provider_detect_platform.hpp`, renaming it into `random_provider_base.hpp` in the process, because the above confuses dependency scanners; they can't see through the macro.

doc/uuid.html
```diff
@@ -591,34 +597,106 @@
...
+    </p>
 
-//for convenience boost::uuids::random_generator
-//is equivalent to boost::uuids::basic_random_generator&lt;boost::mt19937&gt;
+<h5>Preprocessor Macros</h5>
+    <p>
+        <ul>
+            <li>
+                <tt>BOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX</tt>: on Unix this will force the
+                selection of <tt>/dev/*random</tt> over getentropy(3).
```
pdimov: Prefer to not provide this. By all means give people a macro to override behavior *when they ask for it*, no need to to it preemptively.

doc/uuid.html
```diff
@@ -591,34 +597,106 @@
...
+</li>
+            <li>
+                <tt>BOOST_UUID_RANDOM_PROVIDER_FORCE_WINCRYPT</tt>: on Windows this will force the
+                selection of Wincrypt over BCrypt.
```
pdimov: Ditto. No need to ever prefer the older API. If someone asks for it, fine; until then, no need.

doc/uuid.html
```diff
@@ -591,34 +597,106 @@
...
+</li>
+            <li>
+                <tt>BOOST_UUID_RANDOM_PROVIDER_NO_LIB</tt>: disable Windows auto linking.
+            </li>
+            <li>
+                <tt>BOOST_UUID_RANDOM_PROVIDER_POSIX_BLOCKING</tt>: when using <tt>/dev/*random</tt>,
+                if this is defined, skip attempting to open <tt>/dev/urandom</tt> first.
```
pdimov: Nope. Not only should `/dev/random` not be configurable to be used as a preference, it should never be used, anywhere, for anything, as the OpenBSD docs say.

include/boost/uuid/detail/random_provider_bcrypt.ipp
```diff
@@ -0,0 +1,79 @@
...
+
+    //! Obtain entropy and place it into a memory location
+    //! \param[in]  buf  the location to write entropy
+    //! \param[in]  siz  the number of bytes to acquire
+    void get_random_bytes(void *buf, size_t siz)
+    {
+        boost::winapi::NTSTATUS_ status =
+            boost::winapi::BCryptGenRandom(
+                hProv_,
+                reinterpret_cast<boost::winapi::PUCHAR_>(buf),
```
pdimov: `static_cast` is better, when it works, which it does here.

include/boost/uuid/detail/random_provider_detect_platform.hpp
```diff
@@ -0,0 +1,68 @@
...
+//
+// Platform-specific random entropy provider platform detection
+//
+
+#ifndef BOOST_UUID_DETAIL_RANDOM_PROVIDER_PLATFORM_DETECTION_HPP
+#define BOOST_UUID_DETAIL_RANDOM_PROVIDER_PLATFORM_DETECTION_HPP
+
+#include <boost/predef/library/c/gnu.h>
+#include <boost/predef/platform/cloudabi.h>
+// #include <boost/predef/os/bsd/open.h> re-enable when Boost.Predef #66 clears
```
pdimov: No need to wait for Predef. The functions we need are available on OpenBSD 2.5 and above.

include/boost/uuid/detail/random_provider_detect_platform.hpp
```diff
@@ -0,0 +1,68 @@
...
+#  define BOOST_UUID_RANDOM_PROVIDER_WINCRYPT
+#  define BOOST_UUID_RANDOM_PROVIDER_NAME random_provider_wincrypt
+# else
+#  error Unable to find a suitable windows entropy provider
+# endif
+
+#elif BOOST_LIB_C_GNU >= BOOST_VERSION_NUMBER(2, 25, 0) \
+   && !defined(BOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX)
+/* when Boost.Prefef #6 merges, allow this based on OpenBSD documentation:
+BOOST_OS_BSD_OPEN >= BOOST_VERSION_NUMBER(5, 6, 0)) */
```
pdimov: "getentropy() is not intended for regular code; please use the arc4random(3) family of functions instead."

https://man.openbsd.org/getentropy.2

Heed what the OpenBSD people say. They know their stuff. :-)

include/boost/uuid/detail/random_provider_wincrypt.ipp
```diff
@@ -0,0 +1,86 @@
...
+* $Id$
+*/
+
+#include <boost/core/ignore_unused.hpp>
+#include <boost/winapi/crypt.hpp>
+#include <boost/winapi/get_last_error.hpp>
+#include <boost/throw_exception.hpp>
+#include <string>
+
+#if !defined(BOOST_UUID_RANDOM_PROVIDER_NO_LIB)
```
pdimov: BOOST_ALL_NO_LIB, as I already said.

include/boost/uuid/detail/random_provider_wincrypt.ipp
```diff
@@ -0,0 +1,86 @@
...
+{
+  public:
+    random_provider_base()
+        : hProv_(NULL)
+    {
+        if (!
+#if defined(BOOST_NO_ANSI_APIS)
+          boost::winapi::CryptAcquireContextW
+#else
+          boost::winapi::CryptAcquireContextA
```
pdimov: No reason to ever use the A function here, if you ask me.

pdimov:
Ah. It now occurs to me that you need the configuration macros in order to force-test a provider when it wouldn't be selected otherwise. Makes sense.

include/boost/uuid/detail/random_provider.hpp
jeking3: Okay, I will update this.

include/boost/uuid/detail/random_provider.hpp
jeking3: I needed platform detection separate from loading the implementation to make the mocking work properly.  I can expand that out so the macro isn't there and put additional preprocessor branches based on the `BOOST_UUID_RANDOM_PROVIDER_*` definitions made by the detection logic in that header.

doc/uuid.html
jeking3: I'm leveraging this to get coverage and a (mostly) common sad path test.  I'll look at testing against the posix.ipp implementation directly when running on linux.

doc/uuid.html
jeking3: I'm leveraging this to get coverage and a (mostly) common sad path test.  This test is more complex as I have to build a DLL and link against it, but I can make another test file that's specific to the wincrypt.ipp file and loads the mock library.

doc/uuid.html
jeking3: Easy to remove.  Thanks.  Less options too and it simplifies the code and the sad test path that has to account for two open()s in the failure case.

include/boost/uuid/detail/random_provider_bcrypt.ipp
jeking3: Agree, I'll change it.  I usually try static_cast first and let the compiler force me the other way.

include/boost/uuid/detail/random_provider_detect_platform.hpp
jeking3: Good to know, I was planning on getentropy() there so... (on to the next comment)

include/boost/uuid/detail/random_provider_detect_platform.hpp
jeking3: and there you go.

include/boost/uuid/detail/random_provider.hpp
pdimov: Yes, it's better to spell it out with #ifdefs instead of a macro include.

doc/uuid.html
pdimov: Yes, I hadn't yet gotten to the testing part. Keep those macros. I was looking at them from the user side.

include/boost/uuid/detail/random_provider_wincrypt.ipp
jeking3: Sorry I missed that!  Isn't each library capable of having its own preprocessor macros in this area, like `BOOST_SYSTEM_NO_LIB`, `BOOST_CONFIG_NO_LIB`, `BOOST_PYTHON_NO_LIB`...?

jeking3:
That's what I was using them for, however I could write one test per provider implementation instead I suppose.  I just wanted it to be as uniform as possible since they all follow the same pattern.

include/boost/uuid/detail/random_provider_wincrypt.ipp
pdimov: It is, but you have to check the global macro as well and not autolink if it's set. It's a global user switch that requests all autolinking to be disabled, anywhere.

include/boost/uuid/detail/random_provider_wincrypt.ipp
jeking3: I believe you may be right now that the provider string / enumeration stuff is gone.  If so I'll simplify and remove the additional test for this macro.

doc/uuid.html
jeking3: That'll save a bunch of additional work... thanks.

pdimov:
Yes, you could include the implementation header directly and test that, which is still not a bad idea, but the other way is good too, as it can test the component integrated into the rest of the code. So, either way is fine with me, I suppose.

jeking3:
As an aside, we need to get this level of CI integration into every project.

pdimov:
It still seems to me that RtlGenRandom is quite a convenient choice on Windows as it requires no setup and is quite likely a bit faster than firing up a bcrypt/wincrypt provider. I'm not entirely clear on why you dislike it so. :-)

jeking3:
I believe it would still have a setup cost insofar as needing to perhaps statically initialize the address of the API call through LoadLibrary and GetProcAddress, and in a previous PR I was told to avoid using statics.  I agree those two calls are probably much lighter than wincrypt.  I can give it a try.

pdimov:
There's no need for `LoadLibrary`.

```
extern "C" unsigned char __stdcall SystemFunction036( void * p, unsigned long n );
```

works for me.

include/boost/uuid/detail/random_provider.hpp
jeking3: I tried this but it didn't work - I'll keep on it.
```
        typedef Iter::value_type iter_value_type;
        BOOST_STATIC_ASSERT(is_integral<iter_value_type>::value == true_type);
        BOOST_STATIC_ASSERT(is_unsigned<iter_value_type>::value == true_type);
        BOOST_STATIC_ASSERT(sizeof(iter_value_type) * CHAR_BIT >= 32);
```
In reading [SeedSeq documentation](http://www.boost.org/doc/libs/1_65_1/doc/html/boost_random/reference.html#boost_random.reference.concepts.seed_sequence), Iter can be a class with a value_type type definition of an unsigned integral and an operator*, or it can be a pointer to an unsigned integral, correct?  I'm not certain (yet) how I would test for both.

jeking3:
You are assuming that advapi32.dll has already been loaded, and I think that's an unsafe assumption to make.

include/boost/uuid/detail/random_provider.hpp
pdimov: `typedef std::iterator_traits<Iter>::value_type value_type;`

pdimov: Then just `BOOST_STATIC_ASSERT( boost::is_integral<value_type>::value );`, comparing to the type `true_type` doesn't make sense.

pdimov: It will work in C++11 because `true_type` has a `constexpr` conversion to `true` but is still nonsensical. :-)

pdimov:
It works for me in a sample program. Not sure how I could be assuming anything if the program links.

pdimov:
```
extern "C" unsigned char __stdcall SystemFunction036( void * p, unsigned long n );

bool RtlGenRandom( void * p, unsigned long n )
{
	return SystemFunction036( p, n );
}

#include <iostream>

int main()
{
	unsigned x = 0;
	bool r = RtlGenRandom( &x, sizeof(x) );

	std::cout << x << std::endl;
}
```

```
C:\Projects\testbed>g++ testbed.cpp

C:\Projects\testbed>a
725953787

C:\Projects\testbed>clang++ testbed.cpp

C:\Projects\testbed>a
1701069008
```

jeking3:
You are still relying on the fact that ADVAPI32.DLL was loaded.  While it is in the default library list for most build systems, that can be overridden.  ADVAPI32.DLL is not available on Windows CE.  Pretty much everyone will be using bcrypt at this point with wincrypt left to cover older platforms.  Since that code has worked for a very long time here and in Boost.Random, and seeing as to how it's out of the normal code path, doesn't it make sense to leave it as is?

include/boost/uuid/detail/random_provider.hpp
jeking3: I want to support C++03, will it still work?
I don't think I have ever used std::iterator_traits before.
I'm learning a ton of new tricks, thanks!

Lastique:
@pdimov The `LoadLibrary`/`GetProceAddress` approach is what MS suggests [here](https://msdn.microsoft.com/en-us/library/windows/desktop/aa387694(v=vs.85).aspx). All in all, this function seems like one of those semi-legal half-documented APIs that are Windows implementation details and can change at any time. IMHO, we can use such APIs but only if we have to. If there is an official public API that covers our case, we should preferably use it.

jeking3:
Okay that actually wasn't too bad, I'm doing some testing...

pdimov:
It's impossible for a library to not be loaded if you link to it. If that's unsafe, your code is at least 20 times as unsafe.

jeking3:
Would you prefer that I push a delta rather than a full rebase - would that make the next round easier?

pdimov:
> All in all, this function seems like one of those semi-legal half-documented APIs that are Windows implementation details and can change at any time.

I'm willing to bet a considerable amount of money that it won't because approximately all Windows programs presently in existence link to it. And if it does, we'll fix our code. This is not rocket science.

pdimov:
Full rebase will still be better, I think.

include/boost/uuid/detail/random_provider_wincrypt.ipp
jeking3: In my builds on Windows with msvc-14.1, BOOST_ALL_NO_LIB is set to 1 by default.  This seems odd - why would it be set on the way in?  Boost.Uuid does not set it.

pdimov:
On second thought, it probably doesn't matter. You could just as well push another commit here.

Benchmark results including RtlGenRandom will be appreciated.

include/boost/uuid/detail/random_provider_wincrypt.ipp
pdimov: Interesting. No idea why it's set; you'll need to investigate.

Lastique:
On November 9, 2017 3:27:29 AM Peter Dimov <notifications@github.com> wrote:

>> All in all, this function seems like one of those semi-legal 
>> half-documented APIs that are Windows implementation details and can change 
>> at any time.
>
> I'm willing to bet a considerable amount of money that it won't because 
> approximately all Windows programs presently in existence link to it. And 
> if it does, we'll fix our code. This is not rocket science.

I won't argue, I myself am guilty in using some non-public APIs. :)

include/boost/uuid/detail/random_provider_wincrypt.ipp
jeking3: It is defined for all builds of boost, it's in boost.jam.  I will add BOOST_UUID_SOURCE and make it conditional on that as well.

jeking3:
I'm just not comfortable using RtlGenRandom in this context.  If this code moves over to Boost.Random in the future (that's where it really belongs anyway), if someone else wants to take on the potential burden of dealing with the consequences of that in order to squeeze a couple percent points of performance out of it, and in that context it probably makes good sense to be vigilant of things like that, then so be it.

pdimov:
There are no consequences. It just works.

swatanabe:
AMDG

On 11/08/2017 05:46 PM, James E. King, III wrote:
> 
> In my builds on Windows with msvc-14.1, BOOST_ALL_NO_LIB is set to 1 by default.  This seems odd - why would it be set on the way in?  Boost.Uuid does not set it.
> 

  Boost.Build sets BOOST_ALL_NO_LIB globally, since the
library dependencies are managed explicitly.  It's set
in Jamroot and inherited by all Boost subprojects.

In Christ,
Steven Watanabe

pdimov:
Hm. This means that `BOOST_*_NO_LIB` is not the proper mechanism to disable autolinking to non-Boost libraries.

I remember user complaints that their `#define BOOST_ALL_NO_LIB` did not disable autolink to non-Boost libs, but it appears that it cannot be used in such a manner.

The original UUID code just autolinked unconditionally, and I don't remember anyone complaining about it specifically.

So maybe leave it as it was, `#ifndef BOOST_UUID_RANDOM_PROVIDER_NO_LIB`, although the people who'd have problem with autolinking to Windows libraries usually want a global off switch, not to hunt down tens of separate macros.

pdimov:
> but it appears that it cannot be used in such a manner.

Or perhaps not. When using Boost.Build, the Jamfile can just link to the proper library, as it already does now on gcc.

include/boost/uuid/detail/random_provider.hpp
```diff
@@ -0,0 +1,75 @@
...
+    {
+        typedef typename std::iterator_traits<Iter>::value_type value_type;
+        BOOST_STATIC_ASSERT(is_integral<value_type>::value);
+        BOOST_STATIC_ASSERT(is_unsigned<value_type>::value);
+        BOOST_STATIC_ASSERT(sizeof(value_type) * CHAR_BIT >= 32);
+
+        for (; first != last; ++first)
+        {
+            get_random_bytes(&*first, sizeof(*first));
+            *first &= (std::numeric_limits<boost::uint32_t>::max)();
```
pdimov: `uint32_t` is not required to exist, and we already know its `max`: `0xFFFFFFFF`. So just use it.

jeking3: `<boost/cstdint.hpp>` guarantees a proper definition.  Isn't `std::numeric_limits<T>::max` a compile-time constant?  It shouldn't matter.

include/boost/uuid/detail/random_provider_bcrypt.ipp
```diff
@@ -0,0 +1,79 @@
...
+// BCrypt provider for entropy
+//
+
+#include <boost/core/ignore_unused.hpp>
+#include <boost/winapi/bcrypt.hpp>
+#include <boost/winapi/get_last_error.hpp>
+#include <boost/throw_exception.hpp>
+#include <string>
+
+#if defined(BOOST_UUID_SOURCE) || (!defined(BOOST_ALL_NO_LIB) && !defined(BOOST_UUID_RANDOM_PROVIDER_NO_LIB))
```
pdimov: `BOOST_UUID_SOURCE` is by convention set when the source of the library is compiled - which we don't have as we're header-only. So it's not quite the right macro name to use here.

jeking3: Okay I will try to switch to adding the lib in the jamfile.

include/boost/uuid/detail/random_provider_posix.ipp
```diff
@@ -0,0 +1,96 @@
...
+    //! Obtain entropy and place it into a memory location
+    //! \param[in]  buf  the location to write entropy
+    //! \param[in]  siz  the number of bytes to acquire
+    void get_random_bytes(void *buf, size_t siz)
+    {
+        size_t offset = 0;
+        do
+        {
+            ssize_t sz = BOOST_UUID_RANDOM_PROVIDER_POSIX_IMPL_READ(
+                fd_, reinterpret_cast<char *>(buf) + offset, siz - offset);
```
pdimov: static_cast :-)

jeking3:
Do we have any FreeBSD test rigs in the matrix that will verify the happy path for arc4random?

include/boost/uuid/detail/random_provider_wincrypt.ipp
```diff
@@ -0,0 +1,81 @@
...
+#include <boost/winapi/crypt.hpp>
+#include <boost/winapi/get_last_error.hpp>
+#include <boost/throw_exception.hpp>
+#include <string>
+
+#if defined(BOOST_UUID_SOURCE) || (!defined(BOOST_ALL_NO_LIB) && !defined(BOOST_UUID_RANDOM_PROVIDER_NO_LIB))
+#   if defined(_WIN32_WCE)
+#      define BOOST_LIB_NAME "coredll"
+#   else
+#      define BOOST_LIB_NAME "advapi32"
```
pdimov: :-)

include/boost/uuid/detail/random_provider_posix.ipp
jeking3: I got one of them before...

include/boost/uuid/detail/random_provider_wincrypt.ipp
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    //! Obtain entropy and place it into a memory location
+    //! \param[in]  buf  the location to write entropy
+    //! \param[in]  siz  the number of bytes to acquire
+    void get_random_bytes(void *buf, size_t siz)
+    {
+        if (!boost::winapi::CryptGenRandom(hProv_, siz, 
+                    reinterpret_cast<boost::winapi::BYTE_ *>(buf)))
```
pdimov: static_cast :-)

include/boost/uuid/random_generator.hpp
```diff
@@ -31,19 +55,21 @@ class basic_random_generator {
...
+    // if the UniformRandomNumberGenerator is a PseudoRandomNumberGenerator
+    // then it gets seeded by a random_provider.
     basic_random_generator()
         : pURNG(new UniformRandomNumberGenerator)
         , generator
           ( pURNG.get()
           , distribution_type
-            ( (std::numeric_limits<unsigned long>::min)()
-            , (std::numeric_limits<unsigned long>::max)()
+            ( (std::numeric_limits<unsigned int>::min)()
```
pdimov: Not clear why these changes from ulong to uint are needed. Would probably be better to keep the original to cut down on the delta unless there's a good reason to switch.

test/Jamfile.v2
```diff
@@ -7,10 +7,18 @@
 project libs/uuid/test
     : requirements
 
-      # boost::random requires this setting for a warning free build:
+	  # define BOOST_UUID_SOURCE to defeat the default BOOST_ALL_NO_LIB in boost.jam
+	  <define>BOOST_UUID_SOURCE=1
+
+      # BOOST_AUTO_LINK does not work outside of MSVC on Windows so this makes cygwin, mingw-w64 work:
+      <toolset>gcc,<target-os>windows:<linkflags>-lbcrypt
```
pdimov: We could link here regardless of whether the toolset is gcc (there's also clang btw), and this will fix the lack of autolink when using Boost.Build. But the problem is that we don't necessarily know what to link to.

`<linkflags>-lbcrypt` isn't correct though, I think that we need `lib bcrypt ;` and then `<library>bcrypt` here.

test/Jamfile.v2
```diff
@@ -101,4 +108,61 @@ test-suite uuid :
...
+    [ run test_detail_random_provider.cpp 
+      : : :
+        <define>BOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX                  # will force POSIX over getentropy
+        <target-os>windows:<build>no                                    # do not bother running on windows
+      : test_detail_random_provider_happy_posix ]
+
+    [ run test_detail_random_provider.cpp 
+      : : :
+        <define>BOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX                  # will force POSIX over getentropy
+        <define>BOOST_UUID_RANDOM_PROVIDER_POSIX_BLOCKING               # will force /dev/random over /dev/urandom
```
pdimov: No longer necessary.

test/test_entropy_error.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <boost/uuid/entropy_error.hpp>
+#include <string>
+
+int main(int, char*[])
+{
+    using namespace boost::uuids;
+
+    entropy_error err(6, "foo");
+    BOOST_TEST_EQ(6, err.errcode());
+    BOOST_TEST_EQ(std::string("foo"), std::string(err.what()));
```
pdimov: Could use BOOST_TEST_CSTR_EQ here.

include/boost/uuid/detail/random_provider.hpp
pdimov: uint32_t only exists when the platform has a 32 bit type.

Why would anyone prefer to write `(std::numeric_limits<boost::uint32_t>::max)();` instead of `0xFFFFFFFF` is beyond me.

include/boost/uuid/detail/random_provider_bcrypt.ipp
jeking3: I renamed the definition to `BOOST_UUID_FORCE_AUTO_LINK` and kept the same logic.  It does not have to be put into the documentation.

test/Jamfile.v2
jeking3: I will try that and see what it does to the mocks.

include/boost/uuid/detail/random_provider_bcrypt.ipp
pdimov: That's probably easiest, otherwise the selection logic would have to be replicated in the Jamfile which would probably be a challenge. :-)

pdimov: We could link to `advapi32` regardless as everyone links to it anyway, WinCE probably not a concern because I suspect not many use Boost.Build there, but I'm not sure how to decide whether to link to `bcrypt`.

test/Jamfile.v2
jeking3: Is there any way to negate this (`<target-os>windows:<library>bcrypt`), so that I can set it in project requirements so everything links against it by default, but then in a specific mock based test negate it so it does NOT link against bcrypt?

include/boost/uuid/detail/random_provider_bcrypt.ipp
jeking3: These auto-link statements are designed to make it easier for folks consuming Boost.Uuid regardless of the build system.

test/Jamfile.v2
pdimov: I vaguely recall that one could use `<x>y:-<library>bcrypt` to remove it, but I won't be surprised if it doesn't work and I'm imagining things. :-)

pdimov: Turns out I'm not, in this case. http://www.boost.org/build/doc/html/bbv2/overview/targets.html#bbv2.overview.targets.requirements

jeking3: A change as proposed would break, for example, MinGW (32-bit) where bcrypt is not part of the distribution, so the bjam directives would need to be as smart as the preprocessor selection logic.  In fact I bet mingw is already broken, but it hasn't been updated since 2013 so I'm not going to worry about it.  I think I'll keep the auto-link and global gcc/windows directives as-is as they work in CI (well, I'll change it from linkflags to library).

pdimov: Not sure what change would break what, as you're already linking to bcrypt in the gcc case.

pdimov: > In fact I bet mingw is already broken

There are many mingws, so some could be. I tried the one I have here, and it passes the tests except for test_serialization.exe, which crashes. But that may well be Serialization's fault.

pdimov: Appveyor has the plain mingw in C:\mingw, so you could test it in addition to mingw-w64, which you currently have. See f.ex. https://ci.appveyor.com/project/pdimov/system

jeking3: Take a look at my appveyor config; I found that a little while ago and marked some jobs as allowed to fail.  Serialization has a large bug backlog (40 to 50) in trac... I don't know if I reported it.

pdimov: Odd, test_serialization passed the second time. Mystery.

pdimov: Instead of not testing anything, it might be better to disable just test_serialization on mingw and keep testing the rest.

pdimov: Cygwin, on the other hand, passes with flying colors.

jeking3: I am unable to use the syntax: `<target-os>windows:-<library>bcrypt`, or `<target-os>windows:-<define>BOOST_ALL_NO_LIB` - both fail in bjam as an invalid target-os.  I'm going back to the logic I had.

pdimov: If only there existed a simple function in `advapi32` that we could just call and avoid all that hassle.

swatanabe:
AMDG

On 11/08/2017 07:29 PM, James E. King, III wrote:
> jeking3 commented on this pull request.
> 
> 
> 
>> @@ -7,10 +7,18 @@
>  project libs/uuid/test
>      : requirements
>  
> -      # boost::random requires this setting for a warning free build:
> +	  # define BOOST_UUID_SOURCE to defeat the default BOOST_ALL_NO_LIB in boost.jam
> +	  <define>BOOST_UUID_SOURCE=1
> +
> +      # BOOST_AUTO_LINK does not work outside of MSVC on Windows so this makes cygwin, mingw-w64 work:
> +      <toolset>gcc,<target-os>windows:<linkflags>-lbcrypt 
> 
> I am unable to use the syntax: `<target-os>windows:-<library>bcrypt`, or `<target-os>windows:-<define>BOOST_ALL_NO_LIB` - both fail in bjam as an invalid target-os.  I'm going back to the logic I had.
> 

  The '-property' syntax is not very user-friendly.  Basically,
the property that you remove must be exactly the same as the
property that was added:
-<toolset>gcc,<target-os>windows:<linkflags>-lbcrypt

In Christ,
Steven Watanabe

jeking3:
I have some changes to support building for CloudABI compatibility since it's in this pull request, now that I was able to build the Boost.Uuid tests under CloudABI.  These depend on https://github.com/boostorg/predef/pull/68 so I won't push them until that merges.

jeking3:
One of the developers from CloudABI pointed me to arc4random in cloudlibc so I'm going to remove the cloudabi specific implementation and have it use arc4random.

jeking3:
I'm curious what you think about the scope of this change relative to the upcoming release.  I have it marked as a milestone for 1.67.0 and not 1.66.0 right now because it is a lot of change, however it's well tested on every platform, both happy and sad paths.

pdimov:
1.66.0 is closed for beta now so I'd say we missed that boat.

appveyor.yml
```diff
@@ -50,12 +50,12 @@ environment:
...
-  # - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2013
-  #   ADDPATH: C:\mingw\bin;
-  #   TOOLSET: gcc
-  #   CXXFLAGS: cxxflags=-std=c++03
-  #   MAYFAIL: true
-  #   COMMENT: serialization and mingw compatibility issues
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2013
+      ADDPATH: C:\mingw\bin;
+      TOOLSET: gcc
+      CXXFLAGS: cxxflags=-std=c++03
```
pdimov: The new way is `cxxstd=03` instead of using `cxxflags`, which allows testing `cxxstd=03,11` or even `cxxstd=03,11,14,1z` in one go instead of using multiple jobs (slow on Appveyor as they're serialized).

include/boost/uuid/detail/random_provider.hpp
pdimov: 0xFFFFFFFF. :-)

include/boost/uuid/detail/random_provider_wincrypt.ipp
```diff
@@ -0,0 +1,81 @@
...
+#include <string>
+
+#if defined(BOOST_UUID_FORCE_AUTO_LINK) || (!defined(BOOST_ALL_NO_LIB) && !defined(BOOST_UUID_RANDOM_PROVIDER_NO_LIB))
+#   if defined(_WIN32_WCE)
+#      define BOOST_LIB_NAME "coredll"
+#   else
+#      define BOOST_LIB_NAME "advapi32"
+#   endif
+#   define BOOST_AUTO_LINK_NOMANGLE
+#   include <boost/config/auto_link.hpp>
```
pdimov: I've a strong suspicion that these autolinks are unnecessary. Can't imagine "coredll" not being linked on CE, and "advapi32" is on desktop.

pdimov: Hm, nope, a plain `cl` doesn't link `advapi32` by default, so we do need it, unless Winapi autolinks it for us.

pdimov:
This looks good. I however still think that it's way more complex than it would have been with RtlGenRandom on desktop. It's even available by default on Cygwin/Mingw, so things automatically work there even though autolink's not available.

appveyor.yml
jeking3: Sorry, I just re-enabled an old stanza here.  The other ones should be using that.  If not, I'll adjust.

pdimov: All of the jobs at https://ci.appveyor.com/project/jeking3/uuid-gaamf/build/1.0.73-develop seem to be using `cxxflags`.

jeking3: I was thinking about Travis CI, I'll get this fixed; it's queued up behind https://github.com/boostorg/predef/pull/68 right now.

jeking3:
This (hopyfully last) refresh adds changes to support actually running the test suite on cloudabi in a FreeBSD VM.  I switched CloudABI to using arc4random as well.

jeking3:
@swatanabe with one more turn of the crank I think this would be something useful to consider moving into Boost.Random as a replacement for the existing random_device implementation.  Given that the random_device inside Boost.Random today is the only piece of code referencing Boost.System, if we did this then Boost.Random would no longer depend on Boost.System.  I don't know if that would free up Boost.Random from all library based dependencies though.

A fairly simple random_device implementation based on this would be (incomplete...):

```
jking@ubuntu:~/boost/libs/uuid/test$ cat ../include/boost/uuid/detail/random_device.hpp 
#include <boost/uuid/detail/random_provider.hpp>

namespace boost {
namespace uuids {
namespace detail {

template<class Entropy>
class random_device : public random_provider
{
public:
    Entropy operator()()
    {
        Entropy e;
        get_random_bytes(&e, sizeof(Entropy));
        return e;
    }
};
}}}


jking@ubuntu:~/boost/libs/uuid/test$ git diff HEAD test_random_generator.cpp
diff --git a/test/test_random_generator.cpp b/test/test_random_generator.cpp
index bd9566d..6d4e4a0 100644
--- a/test/test_random_generator.cpp
+++ b/test/test_random_generator.cpp
@@ -13,6 +13,7 @@
 #include <boost/detail/lightweight_test.hpp>
 #include <boost/predef/library/c/cloudabi.h>
 #include <boost/random.hpp>
+#include <boost/uuid/detail/random_device.hpp>
 #include <boost/uuid/entropy_error.hpp>
 #include <boost/uuid/random_generator.hpp>
 #include <boost/uuid/uuid_io.hpp>
@@ -117,5 +118,10 @@ int main(int, char*[])
     }
 #endif
 
+    boost::uuids::detail::random_device<unsigned int> randgen;
+    unsigned int i1 = randgen();
+    unsigned int i2 = randgen();
+    BOOST_TEST_NE(i1, i2);
+
     return boost::report_errors();
 }
```

This would improve the platform coverage of random_device to include UWP and CloudABI, and use more optimal calls on newer versions of Linux.  The implementation is also fully tested (happy and sad paths), although the OpenBSD and CloudABI paths are tested manually unless I turn the build into a docker based build.  (I have extensive work doing in the Apache Thrift project, so I know this would be possible).

jeking3:
Should I complete the implementation as a random_device replacement and submit it to Boost.Random?

jeking3:
Now that 1.66.0 is closing out, I want to revisit this, but I need an answer, so please reply:

**Should I complete the implementation as a random_device replacement and submit it to Boost.Random instead?**

The benefits are:

* Broader platform support for entropy generation
* More optimal entropy generation leveraging the latest APIs
* Lower maintenance cost
* Reduced library dependencies (by decoupling random from system)

Thanks - Jim

Lastique:
I though we've been over this and Boost.Random submission failed, haven't we?

My opinion has not changed. You can do it however you like in implementation details of Boost.UUID. A fully compliant implementation of `random_device` in Boost.UUID details seem like an overkill to me; the functionality needed for Boost.UUID can be implemented much more simply. Depending on Boost.Random is only ok if it doesn't require linking with it (which basically means using `random_device` from Boost.Random is not an option).

pdimov:
The right place to ask this question is the list.

jeking3:
I submitted a PR into Boost.Random: https://github.com/boostorg/random/pull/34 - I hadn't heard back from Steven and didn't want to drag the issue up on the mailing list again.

jgzhuang:
@jeking3 Would you mind sharing the details of the "flaws" you mentioned in "Replaced the seed_rng detail implementation (which had a number of flaws) with a replacement header-only random_provider implementation."? 
I did notice a few strange things in the seed_rng in 1.66 and earlier (such as using the uninitialized content, new unsigned int and use the address and delete that unsigned int).  But I am interested in the flaws you discovered. 

Another implementation difference between seed_rng in 1.66 and the random_provider_base in 1.67 is that seed_rng uses std::fread while the random_provider_base uses ::read. With a synthetic testing it seems the latter is slower when siz is 4 although this only affects seeding.

jeking3:
In particular the code was:
 -  relying on uninitialized variables
 -   did not check for errors while obtaining entropy
 - had become difficult to maintain because the entropy generation was not well separated per platform.  
 - was doing way too much to try and be random, reading in process ID, thread ID, time of day, etc. and then running it through a sha1 hash, then pulling some bits.  Much better to use the operating system's entropy generator which was the primary purpose of the change, and to add more platform support.

If you have evidence that shows ::read is slower than std::fread then I suggest opening an issue or a pull request with the evidence and the fix.

Thanks!

PR #54, opened 2017-11-10
revert change to string_generator so it continues to throw std::runtime_error

jeking3:
… like it always had before; per request from @Lastique.


Lastique:
I think you also need to change the docs (uuid.html:504).

Lastique:
Thanks.

PR #55, opened 2018-01-11
Fix relative links in documentation

danieljames:


mclow:
These all look fine to me. I didn't check the links; I'm assuming that they're all correct now.

PR #67, opened 2018-05-07
issue-59: add std::hash support header

### LINKED ISSUE #59: Support std::hash
boost::uuids::uuid now supports only boost:hash via hash_value(), making it an unnecessary dependency on platforms supporting >= c++11. There also should be a std::hash specialization.

jeking3:
This fixes #59


jeking3:
Build failure is from an issue in boost::serialization and unrelated.

PR #74, opened 2018-06-17
Add support for movability to random generators. Fix POSIX random provider

### LINKED ISSUE #71: `boost::uuids::random_generator` is not copyable anymore.
Before Boost v1.67, `boost::uuids::random_generator`  was set to the boost random mtwiter random engine which is copyable. 
In Boost v1.67 (thankfully and finally) it is not set to that random engine but to a new one based on the platform API. Unfortunately it is now not copyable like the previous implementation.

I'm not sure if required to not be copyable but the change can be problematic.
Would it be possible to make it moveable and copyable?

Lastique:
The commit adds support for move constructors and assignment to random UUID
generators and random providers. Also, in POSIX random provider, the commit
improves handling of return value from ::read(), which returns -1
in case of error and may return 0 in case of success. Retry the call when
it was interrupted before reading any bytes from /dev/urandom.

Added new tests for movability.

This fixes #71

### DIFF
--- a/include/boost/uuid/detail/random_provider.hpp
+++ b/include/boost/uuid/detail/random_provider.hpp
@@ -11,10 +11,12 @@
 #ifndef BOOST_UUID_DETAIL_RANDOM_PROVIDER_HPP
 #define BOOST_UUID_DETAIL_RANDOM_PROVIDER_HPP
 
-#include <boost/core/noncopyable.hpp>
+#include <boost/config.hpp>
 #include <boost/cstdint.hpp>
 #include <boost/limits.hpp>
 #include <boost/static_assert.hpp>
+#include <boost/move/core.hpp>
+#include <boost/move/utility_core.hpp>
 #include <boost/type_traits/is_integral.hpp>
 #include <boost/type_traits/is_unsigned.hpp>
 #include <boost/uuid/entropy_error.hpp>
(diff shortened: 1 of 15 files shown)


jeking3:
Interesting that you chose to do this without boost::move, any particular reason?

jeking3:
Ideally we should have a compile-fail test that attempts copy by constructor and another copy by assignment to make sure those are properly disabled.  This would ensure in the future nobody re-enables it by accident.

Lastique:
> Interesting that you chose to do this without boost::move, any particular reason?

I did not want to pull in Boost.Move as a dependency. But if you think it's worth supporting move operations in C++03 then it can be added.

jeking3:
I was trying to do it with Boost.Move but it wasn't cooperating with me.  It was probably me though. :)  I'd prefer it to be implementation that works with both language levels.  The dependencies in move are pretty light:

https://pdimov.github.io/boostdep-report/develop/move.html

I also offered to help the maintainer of Boost.Move add CI to that repository when I found it had none.

Lastique:
I've added support for C++03 and compile-fail tests.

jeking3:
Would be nice to add a unit test for the EINTR handling at some point.  Thanks for the help on move.

PR #75, opened 2018-06-17
Add support for Linux-specific getrandom call to obtain random data.

Lastique:
getrandom is the base implementation for getentropy on Linux. It is also
available in the form of a syscall, which can be called directly on systems with
glibc versions older than 2.25 which don't yet provide wrappers for getrandom or
getentropy but have a recent enough Linux kernel (for example, Debian Stretch).

On systems other than Linux (e.g. Solaris) getentropy is documented as a
source for initialization of a user-space PRNG instead of a direct source
for random data. Since we use the random data directly to initialize UUIDs,
using getrandom on other platforms, where available, would be more preferable
than getentropy. Unfortunately, the only other platform that is known to support
both getentropy and getrandom (Solaris) documents getrandom to return 0
on error, which is a valid return value on Linux. It's not clear if this is
a documentation error or a true incompatibility, and I have no way to test,
so do not use getrandom on Solaris for now. For this reason (and in case
if it is also used on some other platforms) getentropy backend is still
preserved.

### DIFF
--- a/include/boost/uuid/detail/random_provider_getrandom.ipp
+++ b/include/boost/uuid/detail/random_provider_getrandom.ipp
@@ -1,0 +1,77 @@
+//
+// Copyright (c) 2018 Andrey Semashev
+//
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+//   http://www.boost.org/LICENCE_1_0.txt)
+//
+// getrandom() capable platforms
+//
+
+#include <boost/config.hpp>
+#include <boost/throw_exception.hpp>
+#include <boost/predef/library/c/gnu.h>
+#include <cerrno>
+#include <cstddef>
+
+#if !defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_GETRANDOM)
+#if BOOST_LIB_C_GNU >= BOOST_VERSION_NUMBER(2, 25, 0)
+#define BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER
+#elif defined(__has_include)
+#if __has_include(<sys/random.h>)
+#define BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER
+#endif
+#endif
+#endif // !defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_GETRANDOM)
+
+#if defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER)
+#include <sys/random.h>
+#else
+#include <stddef.h> // ssize_t
+#endif
+
+namespace boost {
+namespace uuids {
+namespace detail {
+
+class random_provider_base
+{
+public:
+    //! Obtain entropy and place it into a memory location
+    //! \param[in]  buf  the location to write entropy
+    //! \param[in]  siz  the number of bytes to acquire
+    void get_random_bytes(void *buf, std::size_t siz)
+    {
+        std::size_t offset = 0;
+        while (offset < siz)
+        {
+            ssize_t sz = get_random(static_cast< char* >(buf) + offset, siz - offset, 0u);
+
+            if (BOOST_UNLIKELY(sz < 0))
+            {
+                int err = errno;
+                if (err == EINTR)
+                    continue;
+                BOOST_THROW_EXCEPTION(entropy_error(err, "getrandom"));
+            }
+
+            offset += sz;
+        }
+    }
+
+private:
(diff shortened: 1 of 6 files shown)


include/boost/uuid/detail/random_provider_include_platform.hpp
```diff
@@ -15,6 +15,8 @@
 # include <boost/uuid/detail/random_provider_arc4random.ipp>
 #elif defined(BOOST_UUID_RANDOM_PROVIDER_BCRYPT)
 # include <boost/uuid/detail/random_provider_bcrypt.ipp>
+#elif defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM)
```
jeking3: Was trying to keep these in alphabetical order :)

include/boost/uuid/detail/random_provider_getrandom.ipp
```diff
@@ -0,0 +1,77 @@
...
+            offset += sz;
+        }
+    }
+
+private:
+    static ssize_t get_random(void *buf, std::size_t size, unsigned int flags)
+    {
+#if defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_GETRANDOM)
+        return BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_GETRANDOM(buf, size, flags);
+#elif defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER)
```
jeking3: To implement the unit test could you have the mock implement ::getrandom and have the mock header define BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER?  Unfortunately we're going to have a drop in coverage on the syscall since that (probably) can't be mocked.

include/boost/uuid/detail/random_provider_getrandom.ipp
```diff
@@ -0,0 +1,77 @@
...
+#include <boost/throw_exception.hpp>
+#include <boost/predef/library/c/gnu.h>
+#include <cerrno>
+#include <cstddef>
+
+#if !defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_GETRANDOM)
+#if BOOST_LIB_C_GNU >= BOOST_VERSION_NUMBER(2, 25, 0)
+#define BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER
+#elif defined(__has_include)
+#if __has_include(<sys/random.h>)
```
jeking3: I've never seen __has_include before.  I was just thinking about how nice it would be to have the following include syntax in C++:

```
#include <boost/uuid/sha1.hpp> or <boost/uuid/detail/sha1.hpp>
```

To handle headers that move without leaving behind forwarding headers.

Lastique:
> Is the performance difference between getentropy and getrandom so much that it warrants another code path?

I don't expect performance to be significantly different. The main reasons for this backend are (a) supporting more platforms (glibc < 2.25) and (b) using the more appropriate API for random data generation. As I noted, `getentropy` is mostly targeted for seeding user-space PRNGs. I.e. read some (not much) random data to initialize a UPRNG and then obtain most of the random data from UPRNG. Given that on Linux `getentropy` is based on `getrandom`, this part is mostly puristic, but still worth mentioning. It may be significant on other systems.

> In the test Jamfile you may want to add explicit cases that force both of the linux implementations to execute in positive and negative test cases to ensure test coverage. Without that, most (maybe all) CI builds will test getrandom and nothing will be testing getentropy any more.

I'm not sure it is worth doing. On Linux, `getentropy` backend is not going to be used, ever, so why test it there. I guess, you probably want to "emulate" other systems by forcing `getentropy`?

include/boost/uuid/detail/random_provider_getrandom.ipp
Lastique: I don't like redefining system functions. I've seen you did this with `getentropy` and I deliberately did not follow this route. I think, that mock should be changed as well, but didn't do it in this PR as it was unrelated.

jeking3:
Yes, since we don't have a Solaris CI environment, it's nice to ensure the code doesn't rot somehow.

include/boost/uuid/detail/random_provider_getrandom.ipp
Lastique: >I've never seen __has_include before.

It's one of the SD-6 macros most recent compilers support.

https://en.cppreference.com/w/cpp/experimental/feature_test

Lastique:
I've updated the PR. For tests, I've added the internal `BOOST_UUID_RANDOM_PROVIDER_DISABLE_GETRANDOM` macro, which is only used by tests on Linux, currently. I've also updated the docs.

PR #77, opened 2018-08-15
Disable getrandom provider on Android prior to API version 28.

Lastique:
Android supports getrandom and getentropy functions only since API version 28.

As part of the fix converted Linux/Android platform detection from Boost.Predef
to direct checks of platform-specific macros. This is to work around
Boost.Predef problem described in [1] - depending on header inclusion order
Boost.Predef may indicate Linux or Android. When that problem is fixed we
may change back to Boost.Predef.

Fixes https://github.com/boostorg/uuid/issues/76.

[1]: https://github.com/boostorg/predef/issues/81#issuecomment-413329061

### DIFF
--- a/include/boost/uuid/detail/random_provider_detect_platform.hpp
+++ b/include/boost/uuid/detail/random_provider_detect_platform.hpp
@@ -14,10 +14,15 @@
 #include <boost/predef/library/c/cloudabi.h>
 #include <boost/predef/library/c/gnu.h>
 #include <boost/predef/os/bsd/open.h>
-#include <boost/predef/os/linux.h>
 #include <boost/predef/os/windows.h>
-#if BOOST_OS_LINUX
+
+// Note: Don't use Boost.Predef to detect Linux and Android as it may give different results depending on header inclusion order.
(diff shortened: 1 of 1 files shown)


include/boost/uuid/detail/random_provider_detect_platform.hpp
```diff
@@ -14,9 +14,10 @@
 #include <boost/predef/library/c/cloudabi.h>
 #include <boost/predef/library/c/gnu.h>
 #include <boost/predef/os/bsd/open.h>
-#include <boost/predef/os/linux.h>
 #include <boost/predef/os/windows.h>
-#if BOOST_OS_LINUX
+// Note: Don't use Boost.Predef to detect Linux and Android as it may give different results depending on header inclusion order.
+// https://github.com/boostorg/predef/issues/81#issuecomment-413329061
+#if (defined(__linux__) || defined(__linux) || defined(linux)) && (!defined(__ANDROID__) || (defined(__ANDROID__) && __ANDROID_API__ >= 28))
```
jeking3: Too bad we have to repeat this sequence twice. :|

Lastique:
I've simplified preprocessor checks and removed duplication.

PR #80, opened 2018-10-11
fix test code warning in matrix build: hides previous local declaration

jeking3:


mclow:
This looks fine to me.

PR #87, opened 2018-11-13
Fixed https://github.com/boostorg/uuid/issues/86

BinCaoWR:


include/boost/uuid/detail/md5.hpp
```diff
@@ -293,6 +294,25 @@ class md5
         (dst)[2] = (unsigned char)((src) >> 16); \
         (dst)[3] = (unsigned char)((src) >> 24);
 
+/* 
+ * The md5 results are read 4 bytes at a time in hash_to_uuid() from
+ * basic_name_generator.hpp, so the endian must be considered here
```
jeking3: The RFC specifies that:
'''
   Set octets zero through 3 of the time_low field to octets zero
      through 3 of the hash.
'''

I think the issue could instead be on the "to string" output and not on the generation.

test/test_md5.cpp
```diff
@@ -37,6 +49,9 @@ int main(int, char**)
         hash.process_bytes(expectations[i].data, expectations[i].len);
         boost::uuids::detail::md5::digest_type result;
         hash.get_digest(result);
+        for (boost::uint32_t j = 0; j < 4; ++j) {
```
jeking3: The byte-swapping is incorrect here (for any endian platform). 
There should be a direct memory comparison (memcmp, below) matching the expectations above.

include/boost/uuid/detail/md5.hpp
jeking3: String generation is not the issue, so I am resolving this comment.

test/test_md5.cpp
jeking3: I still believe this comment is correct.  The result of get_digest should be the same in all cases, not conditional based on the platform.  The result is a series of raw bytes and therefore endian-unaware at this level.

jeking3:
Given that test_md5 passes on a big endian system I am going to close this PR.  These changes are unnecessary.

jeking3:
Actually, I think I was wrong in closing this prematurely.  I am going to leave it open until I finish my investigation on big endian.

jeking3:
Okay, I agree when copying out to a char buffer, endianness must be taken into account when copying from the result.  I was able to reproduce the issue in a big-endian multiarch container.  I have consumed this change into that work where I also aligned the sha1 and md5 tests to ensure they are testing at a byte level and not a 32-bit int level, just to be sure.  When I submit a PR for this I will mark it as closing this PR, but you can leave this one open.

jeking3:
I have the code fixed and working on little and big endian systems (producing the same results).  I'm currently working on a Travis CI multiarch docker container solution between boost-ci and jeking3/bdde with a redhat powerpc fedora release 25 image to ensure it stays fixed.

BinCaoWR:
Great. Thanks for your fix. So this PR is no longer needed, we can close it.

PR #95, opened 2019-03-28
Fixes https://github.com/boostorg/uuid/issues/94

### LINKED ISSUE #94: Avoid undefining BOOST_AUTO_LINK_NOMANGLE if it was previously defined.
Undefining BOOST_AUTO_LINK_NOMANGLE, if it was previously defined, breaks desired behavior of all the boost libraries included after uuid.

Suddenly you will get situation where half of headers link against libname.lib and other half to "fully qualified library name".lib.

tomy2105:
Avoid undefining BOOST_AUTO_LINK_NOMANGLE if it was previously defined.

Maybe better solution would be to directly
#pragma comment(lib, "bcrypt.lib")
#pragma comment(lib, "advapi32.lib")
or
#pragma comment(lib, "coredll.lib")

Since <boost/config/auto_link.hpp> seems to be oriented for various logics for linking boost libs and those 3 are not boost libs.

Didn't want to go to change that big myself, so implemented one that still uses auto_link.hpp but solves the problem.

This fixes #94

### DIFF
--- a/include/boost/uuid/detail/random_provider_bcrypt.ipp
+++ b/include/boost/uuid/detail/random_provider_bcrypt.ipp
@@ -19,9 +19,13 @@
 
 #if defined(BOOST_UUID_FORCE_AUTO_LINK) || (!defined(BOOST_ALL_NO_LIB) && !defined(BOOST_UUID_RANDOM_PROVIDER_NO_LIB))
 #   define BOOST_LIB_NAME "bcrypt"
-#   define BOOST_AUTO_LINK_NOMANGLE
-#   include <boost/config/auto_link.hpp>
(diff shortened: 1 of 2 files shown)


ColinChargyBentley:
Hi,
In fact, the same kind of behavior can happen with BOOST_AUTO_LINK_TAGGED or BOOST_AUTO_LINK_SYSTEM enabled. For example, with BOOST_AUTO_LINK_TAGGED enabled, the linking is failing with this error :
LINK : fatal error LNK1104: cannot open file 'libbcrypt-mt-x64.lib'

I think that would be a good idea to avoid using auto_link for non-boost libs.
Regards,
Colin Chargy

PR #98, opened 2019-04-15
Reuse hash_value impl. for std::hash support

brad-anderson:
A temporary string allocation shouldn't be necessary for hashing


brad-anderson:
Sorry for the noise. It was a mistake to use the github editor without testing this. I'll squash and resubmit.

brad-anderson:
Alright, squashed.

PR #99, opened 2019-04-15
Reuse hash_value impl. for std::hash support

brad-anderson:
A temporary string allocation shouldn't be necessary for hashing

Replaces the messy PR #98

### DIFF
--- a/include/boost/uuid/uuid_hash.hpp
+++ b/include/boost/uuid/uuid_hash.hpp
@@ -12,8 +12,6 @@
 #define BOOST_UUID_HASH_HPP
 
 #include <boost/config.hpp>
-#include <boost/container_hash/hash.hpp>
-#include <boost/uuid/uuid_io.hpp>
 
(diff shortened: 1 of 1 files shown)


Lastique:
I think, it'd be better implemented as `hash_value(uuid const&)` overload (which would be implemented in terms of `hash_range` and without `to_string`). `std::hash` specialization then can forward to `hash_value`. The benefit of this approach is that you get integration into Boost.ContainerHash infrastructure and thus automatic support for any containers relying on Boost.ContainerHash.

Lastique:
Also, the implementation should check for `BOOST_NO_CXX11_HDR_FUNCTIONAL` to be compatible with C++03.

Lastique:
Actually, `hash_value` is already implemented. I've made my version in https://github.com/boostorg/uuid/pull/100.

brad-anderson:
Closing in favor of #100

PR #100, opened 2019-04-15
Implement std::hash in terms of hash_value.

Lastique:
Not only it makes the two implementations equivalent, it also fixes the bug
of to_string throwing an exception in the noexcept function. It is also
more efficient as there is no need for the temporary string construction.

Alternative to https://github.com/boostorg/uuid/pull/99.

### DIFF
--- a/include/boost/uuid/uuid_hash.hpp
+++ b/include/boost/uuid/uuid_hash.hpp
@@ -13,7 +13,7 @@
 
 #include <boost/config.hpp>
 #include <boost/container_hash/hash.hpp>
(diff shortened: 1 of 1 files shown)


jeking3:
I'm working on fixing the libc++ failing build through boost-ci.  bootstrap cannot find clang.

jeking3:
There are other boost-ci issues that need to be resolved, I'm working on that as well.

jeking3:
The MSVC 2010-2013 error is caused by a bad appveyor system image.  Switching to their 2015 image will allow it to work, as shown in the boostorg/rational repoisitory.

PR #110, opened 2019-07-25
Fixed issues with linking with external 3rd parties on Windows

ColinChargyBentley:
Hi,
Currently, when enabling BOOST_AUTO_LINK_TAGGED or BOOST_AUTO_LINK_SYSTEM, the build with boost fails because auto_link is used for non boost libraries (Windows libraries) : bcrypt.lib, coredll.lib and advapi32.lib. Boost auto_link seems to be restricted to boost libraries as it constructs library name depending on Boost options. Since these Windows libs does not follow boost naming convention so auto_link shouldn't be used for thoses.

This has already been discussed : https://github.com/boostorg/uuid/pull/95#issuecomment-514135091and https://github.com/boostorg/uuid/issues/94.

Regards,
Colin Chargy

### DIFF
--- a/include/boost/uuid/detail/random_provider_bcrypt.ipp
+++ b/include/boost/uuid/detail/random_provider_bcrypt.ipp
@@ -18,13 +18,11 @@
 #include <boost/throw_exception.hpp>
 
 #if defined(BOOST_UUID_FORCE_AUTO_LINK) || (!defined(BOOST_ALL_NO_LIB) && !defined(BOOST_UUID_RANDOM_PROVIDER_NO_LIB))
-#   define BOOST_LIB_NAME "bcrypt"
-#   if defined(BOOST_AUTO_LINK_NOMANGLE)
-#      include <boost/config/auto_link.hpp>
-#   else
-#      define BOOST_AUTO_LINK_NOMANGLE
-#      include <boost/config/auto_link.hpp>
-#      undef BOOST_AUTO_LINK_NOMANGLE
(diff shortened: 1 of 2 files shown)


ColinChargyBentley:
Hi @jeking3 ,
I found this in boost auto_link : https://github.com/boostorg/config/blob/f7b608155b3396eb33fe9c800f9090f979f0132e/include/boost/config/auto_link.hpp#L99 . I would like to keep the behavior as close as possible to the previous one (in the case where it worked).

I assumed that the comment in https://github.com/boostorg/config/blob/f7b608155b3396eb33fe9c800f9090f979f0132e/include/boost/config/auto_link.hpp#L450 was wrong because it refers to the #ifdef I copied.

What do you think ? Can you see where the behavior changes ?

Regards,
Colin Chargy

MarcelRaad:
Is anyone still working on this or should I create a new PR? It's only this line that's missing:
https://github.com/boostorg/config/blob/a18911902dff03f37582696a13bf60768c183c55/include/boost/config/auto_link.hpp#L103

ColinChargyBentley:
@MarcelRaad You should create a new PR!

MarcelRaad:
Will do, thanks!

mclow:
Should this PR be closed, then?

MarcelRaad:
Yes, as https://github.com/boostorg/config/commit/68b8da3446df6984b74b1bdf6f458d1425498d29 is in master now, I don't this this is needed anymore.

PR #112, opened 2020-03-05
Fix using non-standard library name layout

MarcelRaad:
As `BOOST_AUTO_LINK_NOMANGLE` is checked after `BOOST_AUTO_LINK_TAGGED`
and `BOOST_AUTO_LINK_SYSTEM` in boost/config/auto_link.hpp, using
tagged or system layout would have used the wrong library names for
linking.

This change is based on https://github.com/boostorg/uuid/pull/110, but
adapted to newer changes for clang-cl in
https://github.com/boostorg/config/commit/a18911902dff03f37582696a13bf60768c183c55#diff-e96bc863ddd09e106922c7fbc8f81d1dR103.

### DIFF
--- a/include/boost/uuid/detail/random_provider_bcrypt.ipp
+++ b/include/boost/uuid/detail/random_provider_bcrypt.ipp
@@ -18,13 +18,12 @@
 #include <boost/throw_exception.hpp>
 
 #if defined(BOOST_UUID_FORCE_AUTO_LINK) || (!defined(BOOST_ALL_NO_LIB) && !defined(BOOST_UUID_RANDOM_PROVIDER_NO_LIB))
-#   define BOOST_LIB_NAME "bcrypt"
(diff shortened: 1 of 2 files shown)


Lastique:
Maybe it'd be better to enhance `auto_link.hpp` so that it can be used to link libraries without mangling? Duplicating elaborate compiler checks doesn't look good.

MarcelRaad:
@Lastique Thanks, great idea, and even trivial! I'm letting the tests cycle in my fork at https://github.com/MarcelRaad/config/pull/1 and close this one if it gets accepted.

MarcelRaad:
PR for Config: https://github.com/boostorg/config/pull/325

mclow:
So, the plan is to abandon this change?

MarcelRaad:
If there's nothing unexpected that would lead to the Config PR being rejected, yes. Closing this for now.

PR #115, opened 2020-07-24
Add missing headers

camolezi:
This makes the header be able to build standalone, making possible C++ clang modules builds.


mclow:
I don't think this is quite right. It adds includes (including boost/config.h) _after_ some boost-specific checks. `BOOST_MSVC `.  Seems to me that those includes should go before that.

PR #116, opened 2020-08-13
Add to_chars for writing uuid strings without allocating

gummif:
This is in particular useful for when allocating a std::string is too expensive.

### DIFF
--- a/include/boost/uuid/uuid_io.hpp
+++ b/include/boost/uuid/uuid_io.hpp
@@ -146,23 +146,38 @@ inline wchar_t to_wchar(size_t i) {
 
 } // namespace detail
 
-inline std::string to_string(uuid const& u)
+template<class OutputIterator>
+OutputIterator to_chars(uuid const& u, OutputIterator out)
 {
-    std::string result;
-    result.reserve(36);
-
     std::size_t i=0;
     for (uuid::const_iterator it_data = u.begin(); it_data!=u.end(); ++it_data, ++i) {
         const size_t hi = ((*it_data) >> 4) & 0x0F;
-        result += detail::to_char(hi);
+        *out++ = detail::to_char(hi);
 
         const size_t lo = (*it_data) & 0x0F;
-        result += detail::to_char(lo);
+        *out++ = detail::to_char(lo);
 
         if (i == 3 || i == 5 || i == 7 || i == 9) {
-            result += '-';
+            *out++ = '-';
         }
     }
+    return out;
+}
+
+inline bool to_chars(uuid const& u, char* first, char* last)
+{
+    if (last - first < 36) {
+        return false;
+    }
+    to_chars(u, first);
+    return true;
+}
+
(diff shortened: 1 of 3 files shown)


include/boost/uuid/uuid_io.hpp
```diff
@@ -146,23 +146,32 @@ inline wchar_t to_wchar(size_t i) {
...
         }
     }
+    return true;
+}
+
+inline std::string to_string(uuid const& u)
+{
+    std::string result;
+    result.reserve(36);
+    to_chars(u, std::back_inserter(result), result.capacity());
```
Lastique, line 174: This should probably use `resize` and pass a pointer to the string `data()`.

include/boost/uuid/uuid_io.hpp
```diff
@@ -146,23 +146,32 @@ inline wchar_t to_wchar(size_t i) {
 
 } // namespace detail
 
-inline std::string to_string(uuid const& u)
+template<class OutputIterator>
+bool to_chars(uuid const& u, OutputIterator out, size_t buffer_size)
```
Lastique, line 150: `std::size_t`.

And it's not quite conventional to pass a size along with an output iterator. Doubly so since you are not outputting anything if `buffer_size < 36`.

I think this method should follow the convention of working with iterators, where you pass only the output iterator, which ensures there is enough storage. Another version specifically designed for raw buffers can be provided. That version would accept a raw pointer and a size of the buffer.

gummif: Yes, it's a bit unconventional. Another overload is a good idea.

include/boost/uuid/uuid_io.hpp
gummif: Will do.

include/boost/uuid/uuid_io.hpp
```diff
@@ -146,23 +146,38 @@ inline wchar_t to_wchar(size_t i) {
...
 
         if (i == 3 || i == 5 || i == 7 || i == 9) {
-            result += '-';
+            *out++ = '-';
         }
     }
+    return out;
+}
+
+inline bool to_chars(uuid const& u, char* first, char* last)
```
jeking3, line 167: I'm not a huge fan of return value checking for success... should this perhaps throw something instead like a std::length_error should the buffer size be insufficient?

jeking3 (commented):
One comment about use of return value.

include/boost/uuid/uuid_io.hpp
Lastique: I think a return value is fine for a low level tool like this.

gummif: This was intended to be similar to the API of https://en.cppreference.com/w/cpp/utility/to_chars (non-throwing and non-allocating) but I am open to changing it.

jeking3:
Hey, sorry for the long delay, I've been off in startup-land.

PR #118, opened 2021-02-03
Drone config

sdarwin:
Drone is a continuous integration framework similar to Travis CI. [The C++ Alliance](https://cppalliance.org/) is offering a hosted Drone server for Boost libraries. Please refer to https://github.com/CPPAlliance/drone-ci for more information and instructions.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,0 +1,43 @@
+# Use, modification, and distribution are
+# subject to the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE.txt)
+#
+# Copyright Rene Rivera 2020.
+
+# For Drone CI we use the Starlark scripting language to reduce duplication.
+# As the yaml syntax for Drone CI is rather limited.
+#
+#
+globalenv={'B2_VARIANT': 'variant=release'}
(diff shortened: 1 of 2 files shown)


jeking3:
I'm confused; are we using drone or GitHub Actions going forward?  I see PRs for both.

sdarwin:
Your decision.  You are welcome to use both, which can provide redundancy and more test coverage.  Sometimes, one of the systems will be slow when the other is fast, so you get faster results.

jeking3:
I've been away for a good long time due to startups (and just peeking in to see what I need to do for the 1.77 release)... this repo uses Travis CI and Appveyor through boost-ci.  That repo was an attempt to standardize testing throughout boost and it ended up with some adoption.  Rolling out updates was always a problem however as it required individual attention to each repository using it.

Is there a general consensus about which system the project is moving towards?

I'd prefer to see each of these actually run in CI before I merge anything.  I wouldn't mind switching to GitHub Actions at some point.  Not sure if they support different endianness yet, which is important for testing this library.

sdarwin:
> I'd prefer to see each of these actually run in CI before I merge anything.

In that case, you might copy the files into a fork in your own github account, and test.

PR #120, opened 2021-07-01
Add BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_DISABLE_LIBC_WRAPPER macro

volo-zyko:
Currently there is no way to use `return ::syscall(SYS_getrandom, buf, size, flags);` implementation for this functionality at the end of file without patching it. New macro allows to switch to that branch.


Lastique:
What is the rationale for force-enabling the syscall implementation? If there is a compatibility issue with some libc or something like that, it'd be better to improve the detection logic instead.

volo-zyko:
For example, we target glibc 2.17 which didn't have `getrandom` (it's available only in 2.25+) but we build on Ubuntu 18.04 with glibc 2.27 and we want to force the syscall version.

> it'd be better to improve the detection logic

In my case I don't see how to do this.

Another note, what's the point in having dead code in the implementation?

Lastique:
> Another note, what's the point in having dead code in the implementation?

What you mean by dead code? On a given platform, only one code branch is enabled.

volo-zyko:
> What you mean by dead code?

Sorry, it was a premature conclusion. I needed to get to this code branch.
```cpp
#else
        return ::syscall(SYS_getrandom, buf, size, flags);
#endif
```
But in the setup I described earlier it's impossible or I don't see how to do it. I don't have `BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_GETRANDOM` defined and my glibc version (in the build environment) is greater than 2.25 hence I get `BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER` defined. Consequently my code compiles with:
```cpp
#elif defined(BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_HAS_LIBC_WRAPPER)
        return ::getrandom(buf, size, flags);
#else
```
The resulting binary cannot run on systems with glibc 2.17.

Lastique:
I see. Disclaimer: I'm not the maintainer. Although I personally don't think that what you're doing is correct (IMHO, you should build your code on a platform that matches or older than that it will run on), if the maintainer is willing to support your case, the patch looks reasonable to me. Minor nitpick: I would rearrange the preprocessor conditions to avoid an empty branch.

Lastique:
On the second thought, no. `BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_GETRANDOM` was not really intended to be user-configurable, so `BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_IMPL_SYSCALL` doesn't look like a user-definable either.  There are a number of user-definable `BOOST_UUID_RANDOM_PROVIDER_DISABLE_*` macros that are checked in `random_provider_detect_platform.hpp` (e.g. `BOOST_UUID_RANDOM_PROVIDER_DISABLE_GETRANDOM`). We could introduce a new one, like `BOOST_UUID_RANDOM_PROVIDER_DISABLE_GETRANDOM_LIBC` or something.

Also, the new macro should probably be documented. Although, it looks like, none of the `BOOST_UUID_RANDOM_PROVIDER_DISABLE_*` macros are documented currently, so the docs could be added for all of them separately.

volo-zyko:
> you should build your code on a platform that matches or older than that it will run on

This is one of the options. Actually we did it previously and we've learned its drawbacks the hard way. Building on old systems modern C++ requires building binutils, compiler and whatnot. There are many kinds of incompatibilities. No one normally cares about such things as building modern compiler on ancient systems - glibc 2.17 is from 2013. :-) We've had it enough in the last 6 years or so. Now we want to try a different technique described [here](https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_3.html#SEC25). We hope that building only our code with certain asm directives will relief from the burden of maintaining patches in the toolchain.

Regarding the proposed patch, I just thought that it might be useful to the community.

> On the second thought, no.

I'll take a look at `random_provider_detect_platform.hpp`, thanks.

volo-zyko:
Ok, I took a look at `random_provider_detect_platform.hpp`, specifically at this code snippet:

```cpp
#elif defined(BOOST_UUID_RANDOM_PROVIDER_HAS_GETRANDOM) && !defined(BOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX) && !defined(BOOST_UUID_RANDOM_PROVIDER_DISABLE_GETRANDOM)
# define BOOST_UUID_RANDOM_PROVIDER_GETRANDOM
# define BOOST_UUID_RANDOM_PROVIDER_NAME getrandom

#elif BOOST_LIB_C_GNU >= BOOST_VERSION_NUMBER(2, 25, 0) && !defined(BOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX) && !defined(BOOST_UUID_RANDOM_PROVIDER_DISABLE_GETENTROPY)
# define BOOST_UUID_RANDOM_PROVIDER_GETENTROPY
# define BOOST_UUID_RANDOM_PROVIDER_NAME getentropy

#else
# define BOOST_UUID_RANDOM_PROVIDER_POSIX
# define BOOST_UUID_RANDOM_PROVIDER_NAME posix
```
I do still want to execute `getrandom`-specific code path. If I wanted to switch to POSIX, I had to define `BOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX` (or define both `BOOST_UUID_RANDOM_PROVIDER_DISABLE_GETRANDOM` and `BOOST_UUID_RANDOM_PROVIDER_DISABLE_GETENTROPY`). So, I don't see a point in changing anything here. Instead, I still have to change the `random_provider_getrandom.ipp` but maybe with a renamed macro and with a more nicely arranged code.

Lastique:
> I still have to change the `random_provider_getrandom.ipp` but maybe with a renamed macro and with a more nicely arranged code.

Yes, that's what I meant.

jeking3:
@volo-zyko also you will want to rebase on develop so tests run, and add a test to cover the new BOOST_UUID_RANDOM_PROVIDER_GETRANDOM_DISABLE_LIBC_WRAPPER compile-time branch.

jeking3 (approved):

volo-zyko:
It's not clear what's wrong with [`posix (gcc-4.4, 98,0x, ubuntu-20.04, ubuntu:16.04)`](https://github.com/boostorg/uuid/runs/5147950931?check_suite_focus=true) setup. The log is missing.

jeking3:
I am re-running, we'll see what happens.

jeking3:
I think we missed the train on getting this into 1.79 but after that, this can be merged for 1.80.

PR #127, opened 2022-01-25
ignore test directory for coverage

jeking3:


jeking3:
It looks like the .codecov.yml has to be committed to take effect.

PR #129, opened 2022-02-01
add a multiarch bigendian build job

jeking3:
Some code had trouble with endianness in the past; we used to run a bigendian job in Travis so this restores it.  Specifically, the MD5 code (and the name generator, which uses it) have endian branches.

NOTE: this pull request disables all the other test jobs to speed development of the solution; the final will have all the jobs again.


jeking3:
Will open again once the boost-ci changes go in to make this easier.

PR #138, opened 2022-10-26
Improve generated x86 code for AVX targets

Lastique:
Prefer `vmovdqu` to `vlddqu` on CPUs supporting AVX. `vlddqu` has one extra cycle latency on Skylake and later Intel CPUs and is not merged to the following instructions as a memory operand, which makes the code slightly larger. Legacy SSE3 `lddqu` is still preferred because it is faster on Prescott and the same as `movdqu` on AMD CPUs. It also doesn't affect code size because `movdqu` cannot be converted to a memory operand as memory operands are required to be aligned in SSE.

Closes https://github.com/boostorg/uuid/issues/137.

Also, re-format the test code for MSVC bug 981648, no functional changes.

### DIFF
--- a/include/boost/uuid/detail/config.hpp
+++ b/include/boost/uuid/detail/config.hpp
@@ -1,5 +1,5 @@
 /*
- *            Copyright Andrey Semashev 2013.
+ *          Copyright Andrey Semashev 2013, 2022.
  * Distributed under the Boost Software License, Version 1.0.
  *    (See accompanying file LICENSE_1_0.txt or copy at
  *          https://www.boost.org/LICENSE_1_0.txt)
@@ -36,6 +36,10 @@
 #define BOOST_UUID_USE_SSE41
 #endif
 
+#if defined(__AVX__) && !defined(BOOST_UUID_USE_AVX)
+#define BOOST_UUID_USE_AVX
+#endif
+
 #elif defined(_MSC_VER)
 
 #if (defined(_M_X64) || (defined(_M_IX86) && defined(_M_IX86_FP) && _M_IX86_FP >= 2)) && !defined(BOOST_UUID_USE_SSE2)
@@ -43,6 +47,9 @@
 #endif
 
 #if defined(__AVX__)
+#if !defined(BOOST_UUID_USE_AVX)
+#define BOOST_UUID_USE_AVX
+#endif
 #if !defined(BOOST_UUID_USE_SSE41)
 #define BOOST_UUID_USE_SSE41
 #endif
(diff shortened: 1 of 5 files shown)


Lastique:
@pdimov Peter, MSVC-14.0 fails with ICE in [this CI run](https://ci.appveyor.com/project/jeking3/uuid-gaamf/builds/45184501/job/2f3s59y8ofopqcds#L1648) in `is_contiguous_range` implementation. Could you take a look please?

pdimov:
I'll look into it.

Lastique:
Ping @jeking3.

Lastique:
Rebased and updated the code to prefer movdqu starting with SSE4.1. It doesn't matter on CPUs not supporting AVX, but it is possible that SSE4.1 code will run on a modern CPU that does prefer movdqu to lddqu.

pdimov:
Why not just use `movdqu` everywhere?

Lastique:
Because lddqu is better on NetBurst CPUs. And there's also a workaround for MSVC codegen bug.

pdimov:
> NetBurst CPUs

Really? Have you seen one recently (as in, in the last decade)? :-)

I had a Thinkpad with a P4D I gave away, its battery lasted about half an hour.

pdimov:
It's still a simplification even if we still pretend to care about VS 2008. Not many parts of Boost still work with it, because it's not tested. (I still test msvc-9.0 on Appveyor in old libraries as a matter of habit but that's more of an exception and is not going to last.)

Lastique:
The bug is only fixed in VS2015; VS2013 and before are affected.

Anyway, I've pushed a commit to use movdqu universally, except for the MSVC workaround.

pdimov:
Yes but without the VS2008 path, it's a single ifdef over _ReadWriteBarrier.

Lastique:
Ok, done. I don't want to remove the VS2008 workaround yet.

pdimov:
Here's some discussion on lddqu vs movdqu, for reference: https://community.intel.com/t5/Intel-ISA-Extensions/LDDQU-vs-MOVDQU-guidelines/m-p/1178965

Lastique:
Yeah, I forgot about that discussion, thanks for digging it up. Although it didn't result in a definitive answer - Intel reps didn't comment. In the end I was left with the opinion I had when I started it - use `lddqu` up until AVX, use `vmovdqu` with AVX and later, as before AVX `lddqu` is not worse than `movdqu` and is sometimes better.

pdimov:
I was going to wait until after the release, but I can merge it now if you insist.

Lastique:
A gentle reminder about this PR.

PR #148, opened 2024-04-23
Align `boost::uuid` to 16 bytes

Lastique:
This improves performance of operations on UUIDs as the UUID object now does not(*) cross page boundary. It may also improve code generation by compilers, which can now assume that loads and stores of UUIDs are not as expensive and therefore it may be more preferable to use vector instructions for that.

(*) It is technically possible that the `boost::uuid` object is still not properly aligned under some conditions. For example, on some 32-bit targets the stack is not aligned to 16 bytes, so placing the UUID object on the unaligned stack may result in unaligned UUID. For this reason, this commit does not change the SIMD intrinsics used in `boost::uuid` operations to their aligned equivalents. Using unaligned load/store instructions on actually aligned memory location on modern CPUs has the same performance as using aligned instructions, so the performance benefit is still there.

Closes https://github.com/boostorg/uuid/issues/139.

### DIFF
--- a/include/boost/uuid/uuid.hpp
+++ b/include/boost/uuid/uuid.hpp
@@ -59,6 +59,14 @@ private:
 
 #if !defined(BOOST_UUID_DISABLE_ALIGNMENT)
 
+#if (defined(__SIZEOF_POINTER__) && (__SIZEOF_POINTER__ >= 8)) || \
+    defined(__x86_64__) /* This covers x86-x32 */ || \
+    defined(_M_AMD64) || defined(_M_ARM64)
+            // Align uuid objects to 16 bytes on 64-bit platforms for performance.
+            // On most 32-bit platforms, default stack and dynamic memory alignment is lower than 16
+            // (typically 4 or 8), which makes it problematic to require uuid alignment of 16.
+            BOOST_ALIGNMENT(16)
+#endif
             std::uint64_t align_u64_;
 
 #endif


Lastique:
I've updated the PR to only require alignment of 4 on 32-bit platforms. At least, on Windows, 32-bit processes only get 4-byte aligned stack.

include/boost/uuid/uuid.hpp
```diff
@@ -138,6 +138,16 @@ struct uuid
 
 public:
 
+    // Align uuid objects to 16 bytes on 64-bit platforms for performance.
+    // On most 32-bit platforms, default stack and dynamic memory alignment is lower than 16
+    // (typically 4 or 8), which makes it problematic to require uuid alignment of 16.
+#if (defined(__SIZEOF_POINTER__) && (__SIZEOF_POINTER__ >= 8)) || \
```
jeking3, line 144: How will this interact if the class is placed into another POD using packed alignment?

Lastique: Packed structs are not portable, and the behavior varies from one compiler to another. See [here](https://godbolt.org/z/49aEz1EGo). Basically, gcc and clang on System V ABI override `uuid` alignment, and MSVC respects `uuid` alignment and packs other data members with no explicit alignment requirements.

Lastique:
@pdimov What's the purpose of `test_alignment`? Is it ok to just test that `offsetof % alignof(uuid) == 0`?

pdimov:
It's something that works today.

Lastique:
Do you mean that `alignof(uuid)` is guaranteed (i.e. documented) to be 1? I don't remember seeing this in the docs.

In any case, my question was what was the intended tested feature in this test. IOW, are you testing that `alignof(uuid)` is exactly 1 or are you testing that the compiler is able to align `uuid` in a struct by adding padding? Depending on what is it you're testing, the test could be updated differently, or the PR could be rejected altogether.

pdimov:
I don't think libraries should overalign their public types, sorry.

I tried to align `uuid` like `__uint128_t`, but this type is also overaligned on some platforms, so I left it at `std::uint64_t`.

Maybe one day we'll get `std::uint128_t` which will be a standard integer type and hence not overaligned, but until that happens, uint64_t it is.

PR #149, opened 2024-04-23
Refactor comparison operators on x86

Lastique:
The first commit refactors `operator<` and `operator<=>` on x86 to use the shared implementation.

The second commit restores the previous behavior, when SSE2 implementation was used on x86, even when `uint128_t` is available. This was changed in https://github.com/boostorg/uuid/commit/b3e3a59a9869eb70fb089dbb88b4e29fe7c91c5e.

### DIFF
--- a/include/boost/uuid/detail/uuid_x86.ipp
+++ b/include/boost/uuid/detail/uuid_x86.ipp
@@ -46,7 +46,7 @@ namespace boost {
 namespace uuids {
 namespace detail {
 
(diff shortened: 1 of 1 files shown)


Lastique:
Please, also note that since you switched to `<cstdint>` all `(u)intN_t` need to be qualified with `std::`. This commit takes care of that for `uuid_x86.ipp`.

pdimov:
Please don't put unrelated changes into the same PR.

PR #151, opened 2024-04-24
Add AVX-512/AVX10.1 implementation of `operator<`/`operator<=>`

Lastique:
The new implementation offers about 20% speed improvement over the SSE2 version on i7 12700K, and about 30% over the `uint128_t`-based version.

Also added GitHub CI jobs with AVX-512 enabled to test the new implementation.

### DIFF
--- a/doc/uuid/configuration.adoc
+++ b/doc/uuid/configuration.adoc
@@ -11,18 +11,24 @@ The library does not require building or any special configuration to be used. H
 |Macro |Description
 
 |`BOOST_UUID_NO_SIMD`
-|If defined, disables any optimizations for http://en.wikipedia.org/wiki/SIMD[SIMD]-enabled processors. Generic versions of algorithms will be used instead. This may result in suboptimal performance. By default, optimized algorithms are used, when the library is able to detect the availability of SIMD extensions at compile time.
+|If defined, disables any optimizations for https://en.wikipedia.org/wiki/SIMD[SIMD]-enabled processors. Generic versions of algorithms will be used instead. This may result in suboptimal performance. By default, optimized algorithms are used, when the library is able to detect the availability of SIMD extensions at compile time.
 
 
 |`BOOST_UUID_USE_SSE2`
-|If defined, enables optimizations for http://en.wikipedia.org/wiki/SSE2[SSE2] exstensions available in modern x86 processors.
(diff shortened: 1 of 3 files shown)


pdimov:
Presuming that GHA runners have AVX-512 was certainly optimistic. :-)

Lastique:
> Presuming that GHA runners have AVX-512 was certainly optimistic. :-)

It was worth a try.

I've removed the commit with CI changes.

pdimov:
https://github.com/actions/runner/issues/1069

pdimov:
We're Asciidoc now thanks to @cmazakas (https://github.com/boostorg/uuid/pull/152).

Lastique:
Updated to Asciidoc.

pdimov:
Having this without CI coverage makes me a bit uneasy, although I suppose we'll get an AVX-512 runner every now and then on either GHA or Drone, depending on the phase of the moon.

I can't even test it locally because I don't have AVX-512. :-)

Lastique:
FWIW, you can run the tests locally under [Intel SDE](https://software.intel.com/en-us/articles/intel-software-development-emulator).

pdimov:
Oh, it occurs to me that even if we get an AVX-512 runner by chance, we'll still not test AVX-512 because we won't be enabling it via `instruction-set`.

Would be nice if we can figure something out.

pdimov:
Using `-march=native` will be much easier. Unfortunately `b2` doesn't support that natively. (But then again it doesn't support `instruction-set` at all for MSVC.)

pdimov:
Unrelated, do you by any chance know why `test_time_generator_v1_2` (the one using `std::atomic`) would hang on GCC 13 on GHA? Does something like that happen in the Boost.Atomic tests? (On Drone, only the 32 bit UBSAN run would hang.)

Lastique:
> Unrelated, do you by any chance know why `test_time_generator_v1_2` (the one using `std::atomic`) would hang on GCC 13 on GHA?

No.

I remember there was a similar problem in Boost.Core, but there it was with CMake jobs. I "solved" it by switching from Makefiles to Ninja.

The last few Boost.UUID CI runs were successful though. Does it still happen for you?

> Does something like that happen in the Boost.Atomic tests?

Not last time I checked. Though I'm not using `std::atomic` there. :)

PR #158, opened 2024-07-04
GHA: Remove actions/checkout, fix apt commands, add retries on net errors

Lastique:
GitHub actions/checkout@v3 is no longer working and v4 is incompatible with older Linux versions due to https://github.com/actions/checkout/issues/1590. To fix this, and permanently eliminate the issue of periodic deprecation of actions/checkout, replace it with manual downloads of git snapshots using curl.

Additionally, fixed apt command lines that would incorrectly expand the list of packages to install. Added options to retry on network errors to reduce the probability of spurious CI failures. Also added git checkout parallel jobs to potentially speed up the checkout.

This should fix CI failures with actions/checkout@v3 and eliminate GHA deprecation warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ on:
       - feature/**
 
 env:
+  GIT_FETCH_JOBS: 8
+  NET_RETRY_COUNT: 5
   UBSAN_OPTIONS: print_stacktrace=1
 
 jobs:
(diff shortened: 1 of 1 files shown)


jeking3:
Don't they have an actions@current so you can stay current and risk having a build break the other way?  Seems less likely; I would be surprised if they do not.

Lastique:
`actions@current` means their upgrades will randomly break CI - when they do the upgrade rather than when our maintainer does. Either way, the breakage takes place.

Lastique:
@pdimov @jeking3 Is this PR not going to be accepted? Do I need to update it or it can be just closed?

pdimov:
There's no need for it at the moment, but we might need it in the future when GHA breaks `actions/checkout` entirely on old images.

Lastique:
Closing due to lack of interest.

PR #167, opened 2024-08-22
Fix missing-field-initializers error in GCC 4.8 and 4.9

k3DW:
Failure seen in [unordered#274](https://github.com/boostorg/unordered/pull/274)
[CE link](https://godbolt.org/z/dcYes3PTc)

### DIFF
--- a/include/boost/uuid/uuid.hpp
+++ b/include/boost/uuid/uuid.hpp
@@ -266,7 +266,7 @@ public:
 
     node_type node_identifier() const noexcept
     {
-        node_type node = {};
+        node_type node = {{}};
 
         std::memcpy( node.data(), this->data + 10, 6 );
         return node;


pdimov:
The correct fix is `{{}}`, but why does this warning not appear in the UUID tests? They also use -pedantic.

pdimov:
Oh I see, I'm not using -Werror.

```
gcc.compile.c++ bin.v2/libs/uuid/test/~hdr-decl-uuid_serialize~hpp.test/gcc-4.8/debug/x86_32/cxxstd-11-iso/threading-multi/visibility-hidden/~hdr-decl-uuid_serialize~hpp.o
In file included from ./boost/uuid/uuid_serialize.hpp:11:0,
                 from libs/uuid/test/compile/decl_header.cpp:19:
./boost/uuid/uuid.hpp: In member function 'boost::uuids::uuid::node_type boost::uuids::uuid::node_identifier() const':
./boost/uuid/uuid.hpp:269:27: warning: missing initializer for member 'std::array<unsigned char, 6u>::_M_elems' [-Wmissing-field-initializers]
         node_type node = {};
                           ^
```

k3DW:
> The correct fix is `{{}}`, but why does this warning not appear in the UUID tests? They also use -pedantic.

Corrected, thanks. I'm not sure why this isn't caught in other places. I'm also not sure why this was only caught in my current Unordered PR, and not in any of the other Unordered tests also using this file.

PR #168, opened 2024-09-16
Optimize generic `is_nil`

Lastique:
Reduce the number of generated branches in the generic `is_nil` by using bitwise operations.

Note that a similar optimization is possible for `operator==`, but in that case there may be a slight reduction in performance if UUIDs are mostly unequal, and a slight speedup if they are mostly equal. If you're interested, I can add the change to `operator==` as well.

### DIFF
--- a/include/boost/uuid/detail/uuid_generic.ipp
+++ b/include/boost/uuid/detail/uuid_generic.ipp
@@ -23,7 +23,7 @@ inline bool uuid::is_nil() const noexcept
     std::uint64_t v = detail::load_native_u64( this->data + 0 );
     std::uint64_t w = detail::load_native_u64( this->data + 8 );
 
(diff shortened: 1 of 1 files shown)


pdimov:
Compilers already do this transformation by the way.

Lastique:
Ok, you can close it then.

I just prefer to not rely on compilers, when possible. :)

pdimov:
I'll wait for CE to have 1.86 so that I can look at the codegen (again.)

pdimov:
Now that 1.86 is on CE (somewhat, that is, not on all compilers), I could verify that this optimization is done by approximately all GCC and Clang versions we support. E.g. https://godbolt.org/z/zPoEP6xoj.

Lastique:
Ok.

PR #171, opened 2024-11-11
uuid::data_type: make operator() constexpr

rressi-at-globus:


pdimov:
These are already `constexpr` on `develop` and `master` and should be in 1.87 (currently [beta RC2](https://lists.boost.org/Archives/boost/2024/11/258261.php)).

jeking3:
If that's true, why is there a delta in this PR on boostorg/develop, which is the staging area for things going into releases?

pdimov:
It's probably been created from a stale develop.

jeking3:
Makes sense, forgot about that :)

jeking3:
Double-checked, agree, closing.

PR #176, opened 2025-05-25
fix: Disable libatomic link for windows arm64

rbqvq:
CLANGARM64 does not have libatomic, disable it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,13 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND NOT CMAKE_CXX_COMPILER_FRONTEND_VA
 
 endif()
 
+# CLANGARM64 does not have libatomic, disable it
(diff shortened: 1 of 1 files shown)


pdimov:
Your current `if` only checks for `MINGW`, which covers both GCC and Clang, whereas the comment refers to Clang. Is `libatomic` needed for GCC MinGW/ARM64?

rbqvq:
> OK, we'll keep it this way until a working GCC appears.

How long will it take? Will the build fail if libatomic is not linked?

pdimov:
I meant that I'll merge your PR as-is for now, and we can tweak the condition at a later date when we can test on GCC.

pdimov:
I suspect that we no longer need this after https://github.com/boostorg/uuid/commit/434329f8dc7fcfb69cd7565ab0318d86772cea39. Can you please check?

rbqvq:
I think it should be all right.
But the ci build failed due to other problems, I'll give you the answer when the fix is done. :(

PR #177, opened 2025-07-03
Add RFC-9562 compliant Max UUID (section 5.10)

jeking3:
Section 5.10 of https://datatracker.ietf.org/doc/rfc9562/ defines a "max" uuid which is not present in boost::uuid.

The clang-win build issue is a known issue: https://github.com/actions/runner-images/issues/12435

### DIFF
--- a/include/boost/uuid/max_generator.hpp
+++ b/include/boost/uuid/max_generator.hpp
@@ -1,0 +1,39 @@
+#ifndef BOOST_UUID_MAX_GENERATOR_HPP_INCLUDED
+#define BOOST_UUID_MAX_GENERATOR_HPP_INCLUDED
+
+// Copyright 2025 James E. King III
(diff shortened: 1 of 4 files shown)


pdimov:
I'm not opposed to adding the "max" UUID from the spec; but I'm not sure we want to add a new generator for each constant. We don't have generators for each namespace constant, for instance.

I suppose there was once the idea that the generators could be used in generic code templated by Generator, but even if this proved useful to anyone - which I doubt - we still have no firm grounds for adding generators for each constant.

I think that a better course of action would be to either add the `max_uuid()` function to the main header, or perhaps add a separate `boost/uuid/constants.hpp` header and put it there.

pdimov:
I've added `boost/uuid/constants.hpp`.

@jeking3 would you please consider making a new PR that adds the max UUID there so that authorship is preserved?

PR #179, opened 2025-08-09
Fix mistakes

ivanpanch:
I have corrected a few minor mistakes.


pdimov:
I want some of these changes but not all. E.g. I don't want the semicolon changes, or moving the period outside the parentheses.

pdimov:
I suppose I can cherry-pick the commits I want.

ivanpanch:
For instance, these variants would be correct:

> This is a sentence (see p. 23 for more information).
> This is a sentence. (See p. 23 for more information.)

On the other hand, this is something that I reject:

> This is a sentence (see p. 23.)

pdimov:
OK, let's go with that.

PR #192, opened 2026-01-22
Remove dependencies on Boost.StaticAssert

Lastique:
Boost.StaticAssert has been merged into Boost.Config, so remove the dependency.


pdimov:
This is dependent on https://github.com/boostorg/type_traits/pull/203.

PR #196, opened 2026-07-23
Add RISC-V Vector Extension implementation of `from_chars`

e4d08:
This adds a RISC-V Vector (RVV) implementation of the `from_chars` algorithm, ported from the x86 SIMD implementation. The implementation supports VLEN >= 128 and handles narrow and wide character types (`char`, `char16_t`, `char32_t`).
This PR is heavily inspired by [Andrey Semashev](https://github.com/Lastique)'s work on x86 SIMD implementations.

The performance effect on Spacemit-X60 (riscv64), gcc 16.1.0, in millions of successful `from_chars()` calls per second:

```
Char     | Generic | RVV
=========+=========+================
char     |  4.123  | 11.249 (2.73x)
char16_t |  4.117  |  9.128 (2.22x)
char32_t |  4.301  |  7.754 (1.80x)
```

The unsuccessful parsing case depends on where the error happens, as the generic version may terminate sooner if the error is detected at the beginning of the input string, while the SIMD version performs roughly the same amount of work but faster. Here are some examples for 8-bit character types (for larger types the numbers are more or less comparable):

```
Error              | Generic  | RVV
===================+==========+================
EOI at 35 chars    |   3.782  | 10.109 (2.67x)
EOI at 1 char      | 145.204  | 10.998 (0.08x)
Missing dash at 23 |   6.625  | 11.329 (1.71x)
Missing dash at 8  |  17.294  | 11.575 (0.67x)
Illegal char at 35 |   4.354  | 10.720 (2.46x)
Illegal char at 0  | 106.487  | 11.016 (0.10x)
```

Early errors (EOI at 1 char, illegal char at 0) are significantly slower with RVV — the 1-character input cannot amortize the SIMD setup cost, resulting in a ~13x slowdown. This is consistent with the x86 implementation's behavior, which shows the same pattern.
The missing dash at position 8 case shows a slight regression (0.67x), unlike x86 which achieves a speedup for this scenario. This is likely because the RVV implementation uses two separate expected dash vectors — one per deinterleaved character pair — rather than x86's single shuffled vector. The early error at position 8 does not recoup the overhead of this approach.
Late errors and successful parses all show worthwhile speedups ranging from 1.71x to 2.73x. In general, for inputs where the SIMD path can reach its full processing width, the RVV implementation provides a consistent ~2x improvement over the generic path across all character types.
The test code that was used for benchmarking is the same as the one in PR #186 (with running the successful parse twice for warmup):
[uuid_from_chars_perftest.cpp](https://github.com/user-attachments/files/30310563/uuid_from_chars_perftest.cpp)

Compile and run with:
```
g++ -O3 -march=rv64gc -std=gnu++17 -I. -o uuid_from_chars_perftest_generic uuid_from_chars_perftest.cpp
g++ -O3 -march=rv64gcv -std=gnu++17 -I. -o uuid_from_chars_perftest_rvv uuid_from_chars_perftest.cpp

taskset --cpu-list 1 ./uuid_from_chars_perftest_generic 
taskset --cpu-list 1 ./uuid_from_chars_perftest_rvv
```

### DIFF
--- a/include/boost/uuid/detail/from_chars_riscv.hpp
+++ b/include/boost/uuid/detail/from_chars_riscv.hpp
@@ -402,9 +402,9 @@ const simd_vector128< std::uint8_t > from_chars_simd_constants< T >::mm_lower_id
     {{ 0x01, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }};
 
 
-BOOST_FORCEINLINE vuint8m1_t load_vector128(const simd_vector128< std::uint8_t >& v, std::size_t vl) noexcept
+BOOST_FORCEINLINE vuint8m1_t load_vector128(const simd_vector128< std::uint8_t >& v) noexcept
 {
-    return __riscv_vle8_v_u8m1(v.bytes, vl);
+    return __riscv_vle8_v_u8m1(v.bytes, 16);
 }
 
 template< typename Char, std::size_t Size = sizeof(Char) >
@@ -413,20 +413,18 @@ struct from_chars_simd_load_traits;
 template< typename Char >
 struct from_chars_simd_load_traits< Char, 1u >
 {
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_16(const Char* p, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_16(const Char* p) noexcept
     {
-        return __riscv_vle8_v_u8m1(reinterpret_cast< const std::uint8_t* >(p), vl);
+        return __riscv_vle8_v_u8m1(reinterpret_cast< const std::uint8_t* >(p), 16);
     }
 
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_4(const Char* p, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_4(const Char* p) noexcept
     {
-        (void)vl;
         return __riscv_vle8_v_u8m1(reinterpret_cast< const std::uint8_t* >(p), 4);
     }
 
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_n(const Char* p, unsigned int n, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_n(const Char* p, unsigned int n) noexcept
     {
-        (void)vl;
         return __riscv_vle8_v_u8m1(reinterpret_cast< const std::uint8_t* >(p), n);
     }
 };
@@ -434,25 +432,21 @@ struct from_chars_simd_load_traits< Char, 1u >
 template< typename Char >
 struct from_chars_simd_load_traits< Char, 2u >
 {
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_16(const Char* p, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_16(const Char* p) noexcept
     {
-        (void)vl;
-        std::size_t vl16 = __riscv_vsetvl_e16m2(16);
-        vuint16m2_t v16 = __riscv_vle16_v_u16m2(reinterpret_cast< const std::uint16_t* >(p), vl16);
+        vuint16m2_t v16 = __riscv_vle16_v_u16m2(reinterpret_cast< const std::uint16_t* >(p), 16);
         return __riscv_vnsrl_wx_u8m1(v16, 0, 16);
     }
 
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_4(const Char* p, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_4(const Char* p) noexcept
     {
-        (void)vl;
         vuint16m1_t v16 = __riscv_vle16_v_u16m1(reinterpret_cast< const std::uint16_t* >(p), 4);
         vuint16m2_t v16_w = __riscv_vlmul_ext_v_u16m1_u16m2(v16);
         return __riscv_vnsrl_wx_u8m1(v16_w, 0, 4);
     }
 
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_n(const Char* p, unsigned int n, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_n(const Char* p, unsigned int n) noexcept
     {
-        (void)vl;
         vuint16m2_t v16 = __riscv_vle16_v_u16m2(reinterpret_cast< const std::uint16_t* >(p), n);
         return __riscv_vnsrl_wx_u8m1(v16, 0, 16);
     }
@@ -461,27 +455,23 @@ struct from_chars_simd_load_traits< Char, 2u >
 template< typename Char >
 struct from_chars_simd_load_traits< Char, 4u >
 {
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_16(const Char* p, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_16(const Char* p) noexcept
     {
-        (void)vl;
-        std::size_t vl32 = __riscv_vsetvl_e32m4(16);
-        vuint32m4_t v32 = __riscv_vle32_v_u32m4(reinterpret_cast< const std::uint32_t* >(p), vl32);
+        vuint32m4_t v32 = __riscv_vle32_v_u32m4(reinterpret_cast< const std::uint32_t* >(p), 16);
         vuint16m2_t v16 = __riscv_vnsrl_wx_u16m2(v32, 0, 16);
         return __riscv_vnsrl_wx_u8m1(v16, 0, 16);
     }
 
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_4(const Char* p, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_4(const Char* p) noexcept
     {
-        (void)vl;
         vuint32m1_t v32 = __riscv_vle32_v_u32m1(reinterpret_cast< const std::uint32_t* >(p), 4);
         vuint32m4_t v32_w = __riscv_vlmul_ext_v_u32m1_u32m4(v32);
         vuint16m2_t v16 = __riscv_vnsrl_wx_u16m2(v32_w, 0, 4);
         return __riscv_vnsrl_wx_u8m1(v16, 0, 4);
     }
 
-    static BOOST_FORCEINLINE vuint8m1_t load_packed_n(const Char* p, unsigned int n, std::size_t vl) noexcept
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_n(const Char* p, unsigned int n) noexcept
     {
-        (void)vl;
         vuint32m4_t v32 = __riscv_vle32_v_u32m4(reinterpret_cast< const std::uint32_t* >(p), n);
         vuint16m2_t v16 = __riscv_vnsrl_wx_u16m2(v32, 0, 16);
         return __riscv_vnsrl_wx_u8m1(v16, 0, 16);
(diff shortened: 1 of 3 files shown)


pdimov:
That's interesting but we need a way to test it.

e4d08:
I could add RISC-V tests to the CI, but I need your opinion on the best way of doing it. I consider cross-compiling and running tests under QEMU, but maybe there are better options. If cross-compiling is eligible, let me know and I will implement it.

pdimov:
I don't know of a better way, so we could try that. But please use a separate PR for it.

e4d08:
@pdimov Hello! Testing is ready at #197. Is there anything else needed?

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
+#ifndef BOOST_UUID_DETAIL_FROM_CHARS_RISCV_HPP_INCLUDED
+#define BOOST_UUID_DETAIL_FROM_CHARS_RISCV_HPP_INCLUDED
+
+// Copyright 2025 Andrey Semashev
```
Lastique: The copyright needs to be corrected, as I'm not the author. At least, not the actual RISC V part.

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
...
+    return __riscv_vle8_v_u8m1(v.bytes, vl);
+}
+
+template< typename Char, std::size_t Size = sizeof(Char) >
+struct from_chars_simd_load_traits;
+
+template< typename Char >
+struct from_chars_simd_load_traits< Char, 1u >
+{
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_16(const Char* p, size_t vl) noexcept
```
Lastique: `std::size_t`?. Here and everywhere else.

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
...
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_4(const Char* p, std::size_t vl) noexcept
+    {
+        (void)vl;
+        return __riscv_vle8_v_u8m1(reinterpret_cast< const std::uint8_t* >(p), 4);
+    }
+
+    static BOOST_FORCEINLINE vuint8m1_t load_packed_n(const Char* p, unsigned int n, std::size_t vl) noexcept
+    {
+        alignas(16) std::uint8_t buf[16] = {};
+        detail::memcpy(buf, p, n);
```
Lastique: `#include <boost/uuid/detail/cstring.hpp>` is missing for this.

I'm not at all knowledgeable about RISC-V V, but I'm wondering if this passing through a temporary buffer can be optimized. Is it possible to pass the number of bytes to load as the vector length? Or use a masked load?

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
...
+BOOST_FORCEINLINE void from_chars_simd_core
+(
+    vuint8m1_t chars1, vuint8m1_t chars2, vuint8m1_t chars3,
+    vuint8m1_t const& expected_dashes1,
+    vuint8m1_t const& expected_dashes2,
+    vuint8m1_t const& char_code1_cmp,
+    vuint8m1_t const& char_code2_cmp,
+    vuint8m1_t const& char_code0_sub,
+    vuint8m1_t const& char_code1_sub,
+    vuint8m1_t const& char_code2_sub,
```
Lastique: You are loading these constants by `load_vector128` at the call site, which returns the vector by value.

1. Are you sure these constants need to be loaded this early? They don't depend on the character type or the algorithm. Perhaps, these loads can be delayed until the constants are actually needed. This could simplify `from_chars_simd_core` signature and reduce vector register pressure (if it is a consideration for this code on RVV).
2. If you do need to pass these constants as arguments, do they need to be passed as const references and not by value? Does it not cause unnecessary copying the constants to the stack?

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
...
+}
+
+template< typename Char >
+BOOST_FORCEINLINE from_chars_result< Char > from_chars_simd(const Char* begin, const Char* end, uuid& u) noexcept
+{
+    static_assert(sizeof(Char) == 1u || sizeof(Char) == 2u || sizeof(Char) == 4u, "Boost.UUID: Unsupported input character type for from_chars");
+
+    using char_constants = uuids::detail::from_chars_simd_char_constants< Char >;
+
+    const std::size_t vl = __riscv_vsetvl_e8m1(16);
```
Lastique: I'm wondering if passing `vl` around is necessary. In many places, we know the vector length we want. In fact, in most functions you seem to require 16-byte vectors, the code simply won't produce the correct result if the hardware e.g. only supports 8-byte vectors (which is technically allowed).

Is this implementation supposed to work somehow on CPUs that only support 8-byte vectors? If I understand correctly, `__riscv_vsetvl_e8m1` would return 8 here, and the following code would not produce the correct result (i.e. it would only process parts of the string and populate parts of the UUID).

If such a CPU is not supported then (a) can this entire implementation be disabled via preprocessor checks, so that the generic implementation is used instead? Otherwise, (b) can this code check the returned `vl` and forward to the generic implementation if it is less than 16? (I would prefer (a), if possible.)

If the code *is* supposed to support CPUs with hardware vectors smaller than 16 then, I believe, this call should be `__riscv_vsetvl_e8m2` or something, i.e. enable vector gluing (or whatever is the proper RISC-V term).

In any case, in the following code, where we require a certain vector length, there doesn't seem to be a need to pass `vl` everywhere - we can just directly specify the length we require to every intrinsic. Unless I'm missing something, which is very much possible.

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
...
+        const vbool8_t expected2 = __riscv_vmsgtu_vx_u8m1_b8(expected_dashes2, 0, vl);
+
+        const vbool8_t is_dash1 = __riscv_vmseq_vv_u8m1_b8(chars1, dash_char, vl);
+        const vbool8_t is_dash2 = __riscv_vmseq_vv_u8m1_b8(chars2, dash_char, vl);
+
+        const vbool8_t missing1 = __riscv_vmandn_mm_b8(expected1, is_dash1, vl);
+        const vbool8_t missing2 = __riscv_vmandn_mm_b8(expected2, is_dash2, vl);
+
+        const long first_missing1 = __riscv_vfirst_m_b8(missing1, vl);
+        const long first_missing2 = __riscv_vfirst_m_b8(missing2, vl);
```
Lastique: I feel this may be further optimized. I think, you can get rid of `dash_char` and compare `chars` against `expected_dashes`. You then mask out the comparison results for characters other than at the dash positions anyway.

Another idea is to form a "middle" chars vector from `chars1` and `chars2`, like it is done for x86 [here](https://github.com/boostorg/uuid/blob/659e53385677785620bb70239661c8d105a30d24/include/boost/uuid/detail/from_chars_x86.hpp#L792). Then you could do just one comparison with the dashes vector. Though I don't know if this would win anything performance-wise.

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
...
+    // Check hexadecimal character validity. Values outside 0-15 will have non-zero upper 4 bits.
+    const vuint8m1_t mm_F0 = load_vector128(constants::mm_F0, vl);
+    const vuint8m1_t upper_hi_bits = __riscv_vand_vv_u8m1(upper, mm_F0, vl);
+    const vuint8m1_t lower_hi_bits = __riscv_vand_vv_u8m1(lower, mm_F0, vl);
+
+    const vbool8_t upper_valid = __riscv_vmseq_vx_u8m1_b8(upper_hi_bits, 0, vl);
+    const vbool8_t lower_valid = __riscv_vmseq_vx_u8m1_b8(lower_hi_bits, 0, vl);
+
+    const vbool8_t upper_invalid = __riscv_vmnand_mm_b8(upper_valid, upper_valid, vl);
+    const vbool8_t lower_invalid = __riscv_vmnand_mm_b8(lower_valid, lower_valid, vl);
```
Lastique: Can't it be replaced with `vmsne` (i.e. "not equal")?

include/boost/uuid/detail/from_chars_riscv.hpp
```diff
@@ -0,0 +1,735 @@
...
+
+template< typename T >
+const simd_vector128< std::uint8_t > from_chars_simd_constants< T >::mm_lower_idx_chars3 =
+    {{ 0x01, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }};
+
+
+BOOST_FORCEINLINE vuint8m1_t load_vector128(const simd_vector128< std::uint8_t >& v, std::size_t vl) noexcept
+{
+    return __riscv_vle8_v_u8m1(v.bytes, vl);
+}
```
Lastique: Is this helper really needed? Seems equivalent to the intrinsic.

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: > Is it possible to pass the number of bytes to load as the vector length? Or use a masked load?

Sure, it is possible. All three `load_packed_n` overloads now load directly with the character count as the vector length and narrow in registers.
The caveat is that the lanes above `n` are tail-agnostic, but it is fine as long as this path only runs for inputs shorter than 36 characters.
Also no need for the `cstring.hpp` for now.

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: 1. I believe gcc actually sinks each load to its first use since the core is `BOOST_FORCEINLINE`, but let's move the loads inside for the signature simplicity.
2. Since the call is inlined, both by-value and by-reference compile identically at `-O2`; at `-O0` it looks like
by-value is worse, because each parameter gets its own slot in the vlenb-scaled RVV stack
area, so every access costs a `csrr vlenb` plus address arithmetic and an extra callee-saved
register: 3077 vs 2997 instructions, a 16-byte larger frame, and 7 more whole-register stores.
`const&` only loses for a genuinely non-inlined call which can't happen while the function is `BOOST_FORCEINLINE`.

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: Yes, as stated in the PR this only works with `VLEN >= 128`, thus CPUs that only support 8-byte vectors aren't supported. 
Actually, it looks like existing preprocessor check already required vectors of at least 16 bytes since, per [the specification](https://docs.riscv.org/reference/isa/extensions/vector/_attachments/riscv-v-spec.pdf), 
> The V vector extension requires Zvl128b.

Nevertheless, I've refined the checks to explicitly require `VLEN >= 128`.
As for the `vl` - right, since the algorithm uses exactly 16-byte vectors, looks like configuring `vl` and passing it everywhere is redundant, so I removed it.

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: I loved the idea so yes. `vslidedown(chars1, 8)` + `vslideup(..., chars2, 8)` builds the middle
vector in two instructions, and the check then needs a single comparison, a single `vfirst`, and
position bookkeeping that collapses to `8 + first_missing`. Looks like on gcc 16.2.1 / `rv64gcv` it cuts off plenty amount of instructions.

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: Actually yes, done it.

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: They are in fact equivalent and it gets inlined anyway, but I think explicit inlining is ok so done that.

include/boost/uuid/detail/from_chars_riscv.hpp
Lastique: Ah, sorry, I didn't realize those constants were loaded from the `Char`-dependent constants.

In x86, we're able to pass references to memory via [this operator](https://github.com/boostorg/uuid/blob/659e53385677785620bb70239661c8d105a30d24/include/boost/uuid/detail/simd_vector.hpp#L28), which would compile into memory operands in instructions where these constants were needed, without explicit loads (unless the compiler is smart enough to synthesize the constants, which is not necessarily a good thing in this case). I guess, this is not an option in RVV, and you must always explicitly load the constants, is that right?

Re. passing by reference vs. by value, I see, thanks for checking.

include/boost/uuid/detail/from_chars_riscv.hpp
Lastique: Note that the tail bytes past `n` (and past 4 in `load_packed_4`) must not be actually loaded (as opposed to "loaded but then masked by hardware") as they may reside in a non-mapped or non-readable region (e.g. if the (possibly incomplete) UUID string ends exactly at the end of a mapped page). Does RVV architecture guarantee that the non-loaded bytes in the vector tail won't trigger a page fault or something?

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: > I guess, this is not an option in RVV, and you must always explicitly load the constants, is that right?

That's right, RVV vector types are sizeless in gcc ("RVV type 'vuint8m1_t' does
not have a fixed size"), so `simd_vector`'s conversion operator (gated on
`sizeof(Vector) <= ByteSize`) can't be instantiated for them, and RVV vector instructions take
no memory operands, so the value has to be in a vector register already, with the length as an
explicit operand. Hence the explicit loads.

include/boost/uuid/detail/from_chars_riscv.hpp
e4d08: > Does RVV architecture guarantee that the non-loaded bytes in the vector tail won't trigger a page fault or something?

Yes.

> The _tail_ elements during a vector instruction's execution are the elements past the current
> vector length setting specified in `vl`. The tail elements do not raise exceptions…
> — [V extension specification](https://github.com/riscvarchive/riscv-v-spec/blob/master/v-spec.adoc#54-prestart-active-inactive-body-and-tail-element-definitions)

So a load with `vl = n < 16` cannot access anything past the n-th character.

Lastique:
Looks good to me.

